PR #15, opened 2020-03-16
Fix build with GCC 8.2.0

shdnx:
For some reason, GCC 8.2.0's glibc has the pre-c++11 overloads of `std::string::erase()` which take a non-const iterator. This change fixes building on that compiler, and should also work on later versions (tested with Clang 9.0.1).

### DIFF
--- a/include/toml++/toml_parser_impl.h
+++ b/include/toml++/toml_parser_impl.h
@@ -221,8 +221,8 @@ TOML_IMPL_START
 						recording_buffer.clear();
 					else
 						recording_buffer.erase(
-							recording_buffer.cbegin() + static_cast<ptrdiff_t>(recording_buffer.length() - pop_bytes),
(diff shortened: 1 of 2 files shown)


marzer:
Ooh, thanks @shdnx! Could you also make the same changes in the single-header version, toml.hpp? Either manually, or using the python script (see [CONTRIBUTING](https://github.com/marzer/tomlplusplus/blob/master/CONTRIBUTING.md)).

marzer:
(side note: CircleCI should have run checks against this PR, and the fact that it didn't means I've misconfigured it somehow. Guess I should investigate that.)

shdnx:
@marzer Sorry about that, I fixed it there as well now!

marzer:
Nice work, thanks @shdnx

shdnx:
@marzer Thank you for the quick reaction!

PR #16, opened 2020-04-03
meson: Permit installation of toml++

ximion:
Hi!
Thank you for this great project! This PR adds support to Meson for installing the toml++ headers system-wide and also generates a pkg-config file so other projects can find them easily.
This is very useful if you have a workflow where there is one system copy of code, as used in Linux distributions and a few other distribution schemes.
Cheers,
    Matthias


marzer:
Awesome, thanks @ximion!

PR #46, opened 2020-07-19
Add more VS debug visualizers

Reedbeta:
Hello! I've added debug visualizers for `toml::table`, `toml::array`, and the `parse_result` and `parse_error` types that show up when using the non-exceptions build of the library. These mainly just cut down on visual clutter in the watch window.

Here's what a table looks like with the new visualizers:
image

For comparison, this is what it looked like before:
image

This is what `parse_result` looks like with a successful parse:
image

And this is what it looks like with an error:
image

Hope this is useful to others!

### DIFF
--- a/vs/toml++.natvis
+++ b/vs/toml++.natvis
@@ -50,4 +50,47 @@
 		<DisplayString>{&amp;bytes,s8} ({position})</DisplayString>
 	</Type>
 
+	<Type Name="::toml::table">
+		<DisplayString>{values}</DisplayString>
+		<Expand>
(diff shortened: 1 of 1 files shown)


marzer:
Oh, fantastic! I was planning on doing this eventually, but debugging natvis is such a PITA... very happy for someone to have done that for me, so thanks :)

Reedbeta:
Thanks, happy to help! Natvis is a lot easier now that you can turn on its diagnostic messages and live-edit it during a debugging session. :)

marzer:
also, TIL you can `reinterpret_cast<>` in natvis

PR #50, opened 2020-07-30
fix node::value_or() to act like node::value()

whiterabbit963:


marzer:
Thanks for contributing! Before I can merge this:
1. Please provide at least _some_ explanation/justification for this PR (e.g. I can't tell what it's actually fixing; it appears to do nothing other than make the non-native case slightly slower).
2. Re-generate toml.hpp (as noted in [CONTRIBUTING](https://github.com/marzer/tomlplusplus/blob/master/CONTRIBUTING.md))

whiterabbit963:
Thanks, I'll fix that.
Here is an example showing the problem.
```
auto table = toml::table{{{"value", 10}}};
const toml::node *value = table.get("value");
if(value && value->is_number())
{
    double number = value->value_or<double>(0);
    cout << number << endl;
}
```
The original output was 0, This change will output 10.0.
The old way would see that it was an integer (a native type), then use the integer conversion, but since the type being requested was a double, it would return the default value.
The documentation specifies that node::value_or() is supposed to function like node::value(). This seemed like the simplest solution. Otherwise, you would risk duplicating the check already in node::value
```
		if constexpr (is_natively_one_of<T, std::string, time, date, date_time>)
		{
			if (type() == node_type_of<T>)
				return { this->get_value_exact<T>() };
			else
				return {};
		}
```

whiterabbit963:
I am unable to get the test environment setup right now. However, since I only removed an ```if constexpr``` block and there was no new code, I would expect it to pass. If you need me to run the tests anyway, I can run them tonight.

marzer:
Ah! Good find. You're right, the tests should be unaffected (though I should probably _add_ a new test to cover this logic).

marzer:
@whiterabbit963 One final thing I should have metioned before: Add yourself to the contributors in README.md 😄

Issue #49, opened 2020-07-26
Copying nodes


whiterabbit963:
**Is your feature request related to a problem? Please describe.**
I am currently in the process of replacing a json library with your toml library, and I have encountered this issue a couple times, where I needed to make a copy of a node. In the first case, I was able to do a move. In other cases, I need a copy.
Specifically, I have a table with one element that is an array. I want to be able to copy that array to a second element in that table.

Existing table:
[table]
foo = [1, 2, 3]

In pseudo-code:
auto *array = table.get("foo");
table["bar"] = *array;

**Describe the solution you'd like**
I am pretty that I can accomplish this by manually iterating over the elements in the toml::node.

**Describe alternatives you've considered**
I was surprised to find that copy constructors were deleted.

**Additional context**
Are there, or will there be any plans to include node copying in the future?

bobfang1992:
Why `std::move` is not an option for you?

whiterabbit963:
@bobfang1992 In my example, I want the final table to look like:
[table]
foo = [1, 2, 3]
bar = [1, 2, 3]

In this case, std::move would create undefined behavior, unless foo is properly erased from the table.

I believe I found a solution that seems to compile. toml::table::insert_or_assign()
Doing further testing to ensure it works.

Reedbeta:
As it happens, I've just encountered this issue as well. In my case, I wanted to create a shadow copy of a table, so that I can detect changes potentially made by other code (so that I know when to save the file).

I can understand deleting the copy and assignment operators, since it could be expensive to accidentally copy a large table; but maybe there could be a `deep_copy` method available for when it's needed.

whiterabbit963:
@Reedbeta I thought about something similar via a `clone()` (same idea as `deep_copy()`), but I really do think copy ctors and operator=() provide the simplest mechanics to understand and use.
In my previous example, doing the following feels simpler:
```
auto *array = table.get("foo");
table["bar"] = *array;
```
rather than
```
auto *array = table.get("foo");
table.insert_or_assign("bar", std::move(array->clone()));
```
Having the copy ctors won't prevent possible errors/optimizations like `void foo(toml::table t);` vs `void foo(const toml::table &t);`, but without them, one doesn't have the freedom to choose a particular implementation.

Since toml tends to be more focused on human readability, I would not expect files > 10MB to be common place. For that reason, copy performance should be less concerning than ease of use.

Reedbeta:
[...]

PR #52, opened 2020-08-08
Add implicit conversion operators from node to node_view

Reedbeta:
Hello! I've made a couple of changes to loosen restrictions on the use of `node_view`. Primarily, I added implicit conversion operators from `node` to `node_view` (const and mutable variants). Also, I re-enabled the move assignment operator on `node_view`.

This was motivated by me running into trouble while trying to write code to traverse a table procedurally. Basically, I wanted to do something like this:
```cpp
toml::node_view<const toml::node> child = myTable;
for (auto key : keyPath)
{
    child = child[key];
}
```
where I'm descending into a table using a list of keys in `keyPath`.

In opting to do this by adding implicit conversion operators (as opposed to e.g. adding a constructor on `node_view`, or explicit conversion functions), I was guided by `std::string`, which similarly has an implicit conversion to `std::string_view`. String views are always const, but here I wanted to allow either const or non-const `node_views` so I added both.

I also found that statements like `child = child[key]` above want the move assignment operator, which was disabled. I didn't see any reason for that to be disabled, especially considering that copy assignment is enabled, so I turned it back on.

I'm open to discussion on this if there are other ways you'd prefer to accomplish this, though. 🙂

**Pre-merge checklist**

- [ ] I've updated any affected documentation (_didn't do this as I don't have Doxygen installed_)
    - [ ] Clang 6 or higher
    - [ ] GCC 7 or higher
    - [x] Visual Studio 2019
- [x] I've forever enshrined myself in glory by adding myself to the list of contributors in [README.md](https://github.com/marzer/tomlplusplus/blob/master/README.md)

### DIFF
--- a/include/toml++/toml_node_view.h
+++ b/include/toml++/toml_node_view.h
@@ -63,6 +63,7 @@ TOML_NAMESPACE_START
 			using viewed_type = ViewedType;
 
 		private:
+			friend class TOML_NAMESPACE::node;
 			friend class TOML_NAMESPACE::table;
 			template <typename T> friend class TOML_NAMESPACE::node_view;
 
@@ -88,13 +89,14 @@ TOML_NAMESPACE_START
 			node_view(const node_view&) noexcept = default;
 
 			///// \brief	Copy-assignment operator.
-			TOML_NODISCARD_CTOR
-			node_view& operator= (const node_view&) noexcept = default;
+			node_view& operator= (const node_view&) & noexcept = default;
 
 			///// \brief	Move constructor.
+			TOML_NODISCARD_CTOR
 			node_view(node_view&&) noexcept = default;
 
-			node_view& operator= (node_view&&) noexcept = delete;
+			///// \brief	Move-assignment operator.
+			node_view& operator= (node_view&&) & noexcept = default;
 
 			/// \brief	Returns true if the view references a node.
 			[[nodiscard]] explicit operator bool() const noexcept { return node_ != nullptr; }
--- a/include/toml++/toml_node.h
+++ b/include/toml++/toml_node.h
@@ -819,6 +819,12 @@ TOML_NAMESPACE_START
 			{
 				return do_ref<T>(*this);
 			}
+
+			/// \brief	Creates a `node_view` pointing to this node.
+			[[nodiscard]] operator node_view<node>() noexcept;
+
+			/// \brief	Creates a `node_view` pointing to this node (const overload).
+			[[nodiscard]] operator node_view<const node>() const noexcept;
 	};
 }
 TOML_NAMESPACE_END
--- a/include/toml++/toml_node.hpp
+++ b/include/toml++/toml_node.hpp
@@ -87,6 +87,18 @@ TOML_NAMESPACE_START
 	TOML_MEMBER_ATTR(const) const source_region& node::source()				const noexcept { return source_; }
 
 	#undef TOML_MEMBER_ATTR
+
+	TOML_EXTERNAL_LINKAGE
+	node::operator node_view<node>() noexcept
+	{
+		return node_view<node>(this);
+	}
+
+	TOML_EXTERNAL_LINKAGE
+	node::operator node_view<const node>() const noexcept
+	{
+		return node_view<const node>(this);
+	}
 }
 TOML_NAMESPACE_END
 
--- a/toml.hpp
+++ b/toml.hpp
@@ -2488,6 +2488,9 @@ TOML_NAMESPACE_START
 			{
 				return do_ref<T>(*this);
 			}
+
+			[[nodiscard]] operator node_view<node>() noexcept;
+			[[nodiscard]] operator node_view<const node>() const noexcept;
 	};
 }
 TOML_NAMESPACE_END
@@ -4420,6 +4423,7 @@ TOML_NAMESPACE_START
 			using viewed_type = ViewedType;
 
 		private:
+			friend class TOML_NAMESPACE::node;
 			friend class TOML_NAMESPACE::table;
 			template <typename T> friend class TOML_NAMESPACE::node_view;
 
@@ -4442,12 +4446,12 @@ TOML_NAMESPACE_START
 			TOML_NODISCARD_CTOR
 			node_view(const node_view&) noexcept = default;
 
-			TOML_NODISCARD_CTOR
-			node_view& operator= (const node_view&) noexcept = default;
+			node_view& operator= (const node_view&) & noexcept = default;
 
+			TOML_NODISCARD_CTOR
 			node_view(node_view&&) noexcept = default;
 
-			node_view& operator= (node_view&&) noexcept = delete;
+			node_view& operator= (node_view&&) & noexcept = default;
 			[[nodiscard]] explicit operator bool() const noexcept { return node_ != nullptr; }
 			[[nodiscard]] viewed_type* node() const noexcept { return node_; }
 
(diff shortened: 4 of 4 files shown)


include/toml++/toml_node_view.h
```diff
@@ -88,13 +89,14 @@ TOML_NAMESPACE_START
 			node_view(const node_view&) noexcept = default;
 
 			///// \brief	Copy-assignment operator.
-			TOML_NODISCARD_CTOR
```
Reedbeta, line 91: This looked like a mistake—nodiscard doesn't make much sense on an assignment operator—I figured it was meant to be on the move constructor below.

marzer: Yup, definite copy-pasta.

marzer:
This looks good, though the move assignment operator being deleted was intentional, to solve the "misleading assignment problem" identified in this comment: https://github.com/marzer/tomlplusplus/issues/49#issuecomment-664432777

Not really sure how else to get around that. I'm sure there's some egregious abuse of the C++ type system that could solve it better than my lazy workaround...

marzer:
Re doxygen: you don't need to regenerate the documentation html itself since that's done by CI, but since you've expanded the public interface of node you'd need to add the appropriate doxygen markup to node.h for the conversion operators to appear in the output.

It's probably not really worth it here, though.

marzer:
Hmm, now that I think about it more, the 'misleading assignment' issue could be 'solved' in the short-term just with some explicit caveat documentation. I'll add some next time I do a documentation pass.

That just leaves one thing before I can merge this: I happened to push a small cleanup commit at around the same time as your PR (like an hour earlier I think); can you re-base on it and regen toml.hpp again? Otherwise CI will probably fail when I merge.

Reedbeta:
Sure, I don't mind adding a couple doc lines for the new conversions.

For the "misleading assignment" issue, I have a thought: what if the assignment operators were lvalue-only overloads? I think that ought to prevent temporary node_views from being assigned to. Not sure if that works, but I'll give it a try in a minute here.

marzer:
I tried making them lvalue only and couldn't get it to compile, something about special members not being allowed to specify value category :(

marzer:
You could probably get around that by deleting the assignment operator, creating a proxy type that implicitly converts, and creating an lvalue assignment operator for _that_, though...

Reedbeta:
Hmm, what was the error you got? I just tried it now and it seemed to work, at least in MSVC. I'm on C++20 though, maybe that makes a difference...

marzer:
I can't remember the exact text of it, but "special members" and "value category" were in it. I'm not at a PC currently so can't repro at the moment.

Reedbeta:
It looks like it works on latest gcc, clang, and MSVC, in C++17 mode—just tested them all on godbolt. https://godbolt.org/z/zsGqra

marzer:
Oh, well, that's good news. I was trying to explicitly delete the rvalue overloads, rather than explicitly default the lvalues, so I guess the semantics are a bit different (or I just fucked it up somehow, heh).

Reedbeta:
Alright, I just pushed my updates, including rebasing onto your latest change - how does it look?

marzer:
Looks great. Thanks @Reedbeta!

whiterabbit963:
I think I ran into this, and just opted to use regular nodes, but this is a nicer solution.

Quick question though. Is there a reason to have
```
	TOML_EXTERNAL_LINKAGE
	node::operator node_view<node>() noexcept
	{
		return node_view<node>(this);
	}
```
since node_view is by definition const/read-only anyway?
It would appear that at best, it is a pointless/harmless distinction, and at worst, it is slightly misleading, but I thought I would point it out.

Reedbeta:
You can still use the node() method or the various as() methods to get a
pointer to the underlying node, and that could be const or non-const.

whiterabbit963:
Except that the node_view api has all of those accessor methods returned as const. If node_view is modeled after string_view, that seems to be the "right" thing to do. In addition, if you were able to get a non-const reference to the underlying node, would that be a bug?

Reedbeta:
> the node_view api has all of those accessor methods returned as const

No, it preserves the constness of the node it's viewing. A `node_view<const node>` returns everything as const, but a `node_view<node>` doesn't.

whiterabbit963:
My understanding of view's is that the underlying data is intended to be immutable. The string_view class appears to uphold this intent. The node_view api also appears to uphold this intent as all of the class member functions are marked const.

If by definition a view is only allowed to access the underlying data as const, then there is effectively no difference between node_view<node> and node_view<const node>, as you should not be allowed to modify the node from the view. If you did want to allow mutability to the underlying node, I have to wonder what the design purpose is behind a node_view, when a node has such a similar API.

You could make a case for returning a non-const node_view, if there were methods that changed how the underlying node was viewed. For example, the string_view has a method remove_prefix(), that provides a way to modify the view of the underlying string, but I do not see how this would translate to a view of a node. I also do not see any methods in the API that would apply.

Is there a code sample that wouldn't compile if the following wasn't implemented?
```
	TOML_EXTERNAL_LINKAGE
	node::operator node_view<node>() noexcept
	{
		return node_view<node>(this);
	}
```

marzer:
> My understanding of view's is that the underlying data is intended to be immutable

No, the date is mutable, but the type of it isn't; you can change a value through a node view of a value<>, add an element to an array through a node view of, an array, etc, but you _can't_ re-assign a value node as a table or something.

marzer:
To further illustrate the distinction @whiterabbit963, have a look at the example at the top of the [node_view API docs page](https://marzer.github.io/tomlplusplus/classtoml_1_1node__view.html). Specifically, this line:
```cpp
tbl["products"][0]["keywords"].as_array()->push_back("heavy");
```
`tbl` is non-const, so the first `node_view` retrieved by `["products"]` is a `node_view<const node>`, as are the two sub-views created by `[0]["keywords"]`.  This means that the call to `as_array()` returns a `toml::array*`, _not_ a `const toml::array*`.

marzer:
> If node_view is modeled after string_view

It is in name only. It's more like a `std::optional<node&>`, if such a thing were legal. In a pre-release version of the library it was called `node_ref`, but that's not a perfect fit either. Naming stuff is hard, heh.

whiterabbit963:
Ah, ok, the underlying type is wrapped by a newly constructed non-const object. The example really helped. Thanks for the explanation!

Effectively, you have expanded upon ```std::optional<std::reference_wrapper<node>>```, which is the STL's solution to the issue. 

Naming it ```node_wrapper``` feels the most accurate, if a little verbose. Agreed, naming stuff is not easy.

marzer:
@Reedbeta I have some unfortunate news; I've had to make these operators explicit- while working on a side project I discovered the implicit conversions were causing some problematic interplay with `node_view::operator==` and friends.

To counter the inconvenience I've added additional constructors for node_view so you can make one from a reference or pointer to any node, as well as deduction guides to reduce the syntactic burden:

```cpp
toml::table tbl = toml::parse("foo.toml");
const auto& ctbl = tbl;
auto ptbl = &tbl;
auto cptbl = &ptbl;

auto nv   = node_view{ tbl }; // node_view<node>
auto cnv  = node_view{ ctbl }; // node_view<const node>
auto nv2  = node_view{ ptbl }; // node_view<node>
auto cnv2 = node_view{ cptbl }; // node_view<const node>
```

Reedbeta:
Ahh, okay. That should be fine for me, thanks for the heads-up.

I suspect the (explicit) conversion operators are a bit redundant now that the new constructors exist? I'm not sure there's any case the conversions would be useful where you couldn't use the constructors as well. In any case, I can easily adjust the couple places in my code where I use them.

marzer:
Yeah, they are definitely a bit redundant now. I considered removing them while I was there but then decided against it mainly because they aren't doing any harm. My gut feeling is that there's some specific language-lawyer-ish reason they might be useful, but I can't think of one myself.

PR #60, opened 2020-09-17
meson: Install CMake and pkg-config files into architecture-independe…

### LINKED ISSUE #59: CMake and pkg-config files should be installed into architecture-agnostic directories
## Environment
**toml++ version and/or commit hash:**  bf26a88412a8547c2fc29819927ad9030dcd66d9

## Describe the bug
CMake and pkg-config files are installed into architecture-specific directories despite the library being header-only and thus always architecture-agnostic.

CMake files should go into `/usr/lib/cmake/tomlplusplus`, pkg-config into `/usr/share/pkgconfig`.

tambry:
…nt directories

toml++ is header-only so these should be installed into architecture-independent directories to allow them to be found for crosscompiling regardless of the architecture they were installed on.

Fixes #59.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -164,7 +164,7 @@ if is_msvc or is_icc_cl
 		'/GF', # string pooling
 		'/openmp-',
 		'/permissive-',
-		'/sdl-', 
+		'/sdl-',
 		'/utf-8',
 		'/volatile:iso',
 		'/Zc:__cplusplus',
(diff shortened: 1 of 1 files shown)


marzer:
Oh, that was fast, thanks @tambry! Want to add yourself to the contributors list in README.md before I merge it?

tambry:
> Oh, that was fast, thanks @tambry! Want to add yourself to the contributors list in README.md before I merge it?

I'd rather not add myself for such a small change. Feel free to go ahead and merge. 🙂

PR #63, opened 2020-09-30
Rename Windows.h to windows.h for mingw-w64

rezahousseini:
Use lower case header file names for compilation with mingw-w64.


marzer:
Ah, thanks!

PR #77, opened 2020-12-16
Make iterators real iterators

HazardyKnusperkeks:
They were missing the iterator_category and thus could not be used with
some standard algorithms.

**What does this change do?**

Making table_iterator and array_iterator usable with standard algorithms and iterator functions.

**Pre-merge checklist**

- [X] I've read [CONTRIBUTING.md]
- [X] I've rebased my changes against the current HEAD of `origin/master` (if necessary)
- [X] I've added new test cases to verify my change
- [X] I've regenerated toml.hpp ([how-to])
    - [ ] Clang 6 or higher
    - [X] GCC 7 or higher

### DIFF
--- a/include/toml++/toml_array.h
+++ b/include/toml++/toml_array.h
@@ -38,6 +38,7 @@ TOML_IMPL_NAMESPACE_START
 			using reference = value_type&;
 			using pointer = value_type*;
 			using difference_type = ptrdiff_t;
+			using iterator_category = typename std::iterator_traits<raw_iterator>::iterator_category;
 
 			array_iterator() noexcept = default;
 			array_iterator(const array_iterator&) noexcept = default;
--- a/include/toml++/toml_common.h
+++ b/include/toml++/toml_common.h
@@ -16,6 +16,7 @@ TOML_DISABLE_WARNINGS
 #include <cstring>
 #include <cfloat>
 #include <climits>
+#include <iterator>
 #include <limits>
 #include <memory>
 #include <string_view>
--- a/include/toml++/toml_table.h
+++ b/include/toml++/toml_table.h
@@ -75,6 +75,8 @@ TOML_IMPL_NAMESPACE_START
 			using value_type = table_proxy_pair<IsConst>;
 			using reference = value_type&;
 			using pointer = value_type*;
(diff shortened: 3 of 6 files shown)


tests/user_feedback.cpp
```diff
@@ -5,6 +5,8 @@
 
 #include "tests.h"
 
+#include <algorithm>
```
HazardyKnusperkeks, line 8: Here I don't know where you would put that include.

marzer: That's the perfect place for it. Just wrap it in warning-disabling bookends:

```suggestion
TOML_DISABLE_WARNINGS
#include <algorithm>
TOML_ENABLE_WARNINGS
```

tests/user_feedback.cpp
```diff
@@ -142,5 +144,55 @@ TEST_CASE("user feedback")
 		using namespace toml::literals; // should compile without namespace ambiguity
 		auto table = "[table]\nkey=\"value\""_toml;
 	}
+
+	SECTION("Using std::distance, std::count_if, etc. with Iterators") // No issue
```
marzer, line 148: I'd prefer if the tests in this file were derived exclusively from user bug reports and feature requests (i.e. where there's a preceding dialogue with a user that I can link back to), so `user_feedback.cpp` perhaps isn't the place for it.

I'd suggest moving the table checks to `manipulating_tables.cpp` and the array checks to `manipulating_arrays.cpp`. Or create an iterator-specific test TU... I'm happy with either.

tests/user_feedback.cpp
marzer: This test code probably doesn't really belong in this file (see my other note), so once it has a new home just wrap it in warning-disabling bookends:

```suggestion
TOML_DISABLE_WARNINGS
#include <algorithm>
TOML_ENABLE_WARNINGS
```

marzer (changes requested):
Thanks for filling this gap! Just need to move things around a little.

tests/user_feedback.cpp
HazardyKnusperkeks: Yeah I thought about that. But my conclusion was I was just not reporting the bug before fixing it. :)

marzer (approved):
Looks great, thanks!

PR #80, opened 2021-01-06
Don't insert emty table to empty table arrays

marshevms:
Doesn't create empty table in empty arrays of tables

    - [ ] Clang 6 or higher
    - [x] GCC 7 or higher

### DIFF
--- a/include/toml++/toml_parser.hpp
+++ b/include/toml++/toml_parser.hpp
@@ -2591,10 +2591,7 @@ TOML_IMPL_NAMESPACE_START
 							).first->second->ref_cast<array>();
 						table_arrays.push_back(tab_arr);
 						tab_arr->source_ = { header_begin_pos, header_end_pos, reader.source_path() };
-						
-						tab_arr->elements.emplace_back(new toml::table{});
-						tab_arr->elements.back()->source_ = { header_begin_pos, header_end_pos, reader.source_path() };
-						return &tab_arr->elements.back()->ref_cast<table>();
+						return parent;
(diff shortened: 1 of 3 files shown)


marshevms:
I think this is not a very elegant solution, but you can see what needs to be done to solve the problem.

marzer:
I'm sorry but what you've identified as a problem is actually by design. Since a TOML `[[array-of-tables]]` is just syntactic sugar for an array containing tables, when one is encountered in the TOML the parser must create both the outer array _and_ the inner table nodes to ensure the document is structurally correct (thus is structurally round-trippable).

That is, given the following TOML:

```toml
a = { "key" = 1 } # inline table
b = []            # array value
[[c]]             # array-of-tables with a single, empty table element
```

And the following C++ to read it in and re-serialize it:
```cpp
const auto tbl = toml::parse(toml_text);
std::cout << tbl << "\n";
```

You should see the following output:
```
a = { key = 1 }
b = []

[[c]]
```

But with your changes, we see this instead:
```
a = { key = 1 }
b = []
c = []
```

Note that the last line of the output TOML is now incorrect, emitting `c = []` instead of `[[c]]`.

marzer:
I do agree that it's a bit wasteful on the C++ side of things, since it adds unnecessary nodes to the TOML tree in cases were people have added empty `[tables]` and `[[arrays-of-tables]]` to the document. I'm happy to consider other solutions to this. How about something like a `prune()` method to remove all the empty leaves from the tree?

PR #88, opened 2021-02-21
Implement configurable precision 

vaartis:
TOML++ currently forces precision to always be `std::numeric_limits<T>::digits10 + 1`, by using `std::setprecision` in its code, which overrides any precision set by user. The reason for the following change is that numbers like 0.3 become unnecessarily long with the default precision settings of 11 symbols. Since the standard already has a way to configure precision, I figured the best way to approach this is to allow to simply disable TOML++'s precision override and let users pass it to the stream themselves.

**What does this change do?**

Introduces a preprocessor option called `TOML_FORMAT_SETPRECISION`, which is enabled by default so as to make the old behaviour the default and not break anything. When this is set to `0`, TOML++ will not force precision and will allow the user to set their own.

No

**Pre-merge checklist**

- [x] I've added new test cases to verify my change - no need, probably
- [x] I've rebuilt and run the tests with at least one of - again, there should be no need for this
    - [x] Clang 6 or higher
    - [ ] GCC 7 or higher


marzer:
Thanks for the contribution but I don't think adding this is a good idea. Even though you've done the sensible thing and made this new behaviour opt-in, I'd suggest a better way to handle this in your own code is to just print the contained value directly (using `.get()` or similar).

The reason the library behaves this way is for round-trippability. Allowing a fallback to the standard's default precision of 6 (or, indeed, allowing _any_ user-defined precision) just opens me up to bug reports about lost information when re-reading a previously written config file. Instead it's far more maintainable to just avoid the stream formatting state altogether.

Other minor points:
- This change won't fall back to the precision set in the user's stream at all, but will instead fall back to the default precision of the local `std::ostringstream` created in the block inside `print_floating_point_to_stream()`.
- It seems this new feature applies only to the stringstream branch; there's a branch to select for the use of `std::to_chars` from `<charconv>` where available, too, and that would still print at full precision regardless of the setting of `TOML_FORMAT_SETPRECISION`.

vaartis:
> I'd suggest a better way to handle this in your own code is to just print the contained value directly (using .get() or similar).

How do I do this if I'm printing the whole table? This is a problem that exists when there's a float value in the table, I don't think there's a way to output everything else up to `key =` and then output the float value in the precision that doesn't expose the flaws of IEEE 754 and make the number far too long even though I just passed 0.3 in.

marzer:
You'd have to do it recursively using a hand-rolled solution unfortunately. I sympathize that it's not ideal if you want "pretty" floats.

vaartis:
Exactly. But this sounds like a very bad solution. I'd rather solve this problem properly somehow. Maybe introducing a variable into the formatter that controls precision would be better? I'd really like to solve this somehow because I'm probably not the only person who wants this (though there haven't been any reports about this apparently so I can't know).

marzer:
If you really want to implement a solution to this then yes adding something to the formatters would be the right move. You might consider determining the precision on the fly by doing a log10 on the integral and rational components of each float (or something along those lines), instead of making it user configurable (which I'm against for the reasons I outlined above, and won't accept any PR which does that).

You'd have to pass that precision into the print method, which would mean updating it for the charconv branch, too.

vaartis:
I don't know if a PR can be reopened, but I have pushed a change that, for floating pointer numbers that have <3 positions after the floating point, prints them with just those, and otherwise it prints them with the usual `std::numeric_limits<T>::digits10 + 1` precision.

marzer:
> I don't know if a PR can be reopened

Ordinarily they can, except in the specific case that a PR is closed _then_ force-pushed to, for some reason. Not sure why ¯\_(ツ)_/¯. Happy for you to submit another PR, sounds like it's heading in the right direction.

vaartis:
Continuation in #89

PR #89, opened 2021-02-22
Introduce a heuristic to print floating point numbers more coincely

vaartis:
**What does this change do?**

Introduces a heuristic that, for floating pointer numbers that have <=3 positions after the floating point, prints them with just those, and otherwise it prints them with the usual `std::numeric_limits<T>::digits10 + 1` precision.

Continuation of #88 

**Pre-merge checklist**

    - [ ] Clang 6 or higher
    - [ ] GCC 7 or higher

### DIFF
--- a/include/toml++/toml_print_to_stream.h
+++ b/include/toml++/toml_print_to_stream.h
@@ -222,12 +222,31 @@ TOML_IMPL_NAMESPACE_START
 					return true;
 				};
 
+				static constexpr auto num_decimal_places = [](T number) {
+					// This prints numbers with up to 3 decimal places more coincely
+					constexpr double precision { 1e-4 };
+
+					T temp {};
+					int decimal_places {};
+					do {
+						number *= 10;
+						temp = number - int(number);
+
+						decimal_places++;
+					} while(temp > precision && decimal_places < std::numeric_limits<T>::digits10 + 1);
+
+					return decimal_places;
+				};
+
+				auto decimal_places = num_decimal_places(val);
+
+
 				#if TOML_FLOAT_CHARCONV
 				{
 					char buf[charconv_buffer_length<T>];
 					const auto res = hexfloat
-						? std::to_chars(buf, buf + sizeof(buf), val, std::chars_format::hex)
-						: std::to_chars(buf, buf + sizeof(buf), val);
+						? std::to_chars(buf, buf + sizeof(buf), val, std::chars_format::hex, decimal_places)
+						: std::to_chars(buf, buf + sizeof(buf), val, std::chars_format::general, decimal_places);
 					const auto str = std::string_view{ buf, static_cast<size_t>(res.ptr - buf) };
 					print_to_stream(str, stream);
 					if (!hexfloat && needs_decimal_point(str))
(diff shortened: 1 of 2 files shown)


vaartis:
I guess I do need to fix the tests somehow. I can't buld them because meson says `unknown target 'tests/cpp17'``...

include/toml++/toml_print_to_stream.h
```diff
@@ -222,12 +222,31 @@ TOML_IMPL_NAMESPACE_START
...
+				auto decimal_places = num_decimal_places(val);
+
+
 				#if TOML_FLOAT_CHARCONV
 				{
 					char buf[charconv_buffer_length<T>];
 					const auto res = hexfloat
-						? std::to_chars(buf, buf + sizeof(buf), val, std::chars_format::hex)
-						: std::to_chars(buf, buf + sizeof(buf), val);
+						? std::to_chars(buf, buf + sizeof(buf), val, std::chars_format::hex, decimal_places)
```
marzer, line 248: You don't need to do this in the hexfloat case; full precision is assumed for hexfloats (that's sorta the whole point of them).

marzer (changes requested):
Also need to regen `toml.hpp` ^_^

vaartis:
After looking at the tests, I'm not sure this is possible to do while still retaining the round-trip thing, because "0.3" is not written as "0.30000000000000004" anymore, basically. I guess this isn't going to work out after all and I'll have to maintain a fork with precision setting disabled after all. Feel free to close this, I dont' really understand how to fix this issue.

marzer:
> I dont' really understand how to fix this issue.
> 
> Hence this change breaks all the tests, basically. Sorry for wasting your time.

That's alright. It's kinda a damned-either-way scenario, really. Thanks for giving it a crack.

marzer:
@vaartis if you would still like this feature I can probably implement it trivially as part of the [v3 branch](https://github.com/marzer/tomlplusplus/tree/v3), since I now use the new `format_flags` for lots of other opt-in formatting stuff. Something like `format_flags::relaxed_float_precision`. Sound good?

vaartis:
> @vaartis if you would still like this feature I can probably implement it trivially as part of the [v3 branch](https://github.com/marzer/tomlplusplus/tree/v3), since I now use the new `format_flags` for lots of other opt-in formatting stuff. Something like `format_flags::relaxed_float_precision`. Sound good?

Please do! Then I can finally remove that fork I've been using with a single line changed..

marzer:
@vaartis I've implemented this in the v3 branch.

```toml
val1 =   0.010284358729827818
val2 =   0.0102
val3 =  10.0102
val4 =  10.010284358729827818
val5 =  10.0
```

```cpp
using formatter = toml::toml_formatter;
std::cout << cfg << "\n\n";
std::cout << formatter{ cfg, formatter::default_flags | toml::format_flags::relaxed_float_precision };
```

```sh
val1 = 0.010284358729827818
val2 = 0.0102
val3 = 10.0102
val4 = 10.010284358729828
val5 = 10.0

val1 = 0.0102844
val2 = 0.0102
val3 = 10.0102
val4 = 10.0103
val5 = 10.0
```

vaartis:
Thank you for implementing this. I think it's probably enough for it to be per-stream, for my use-case it certainly should be. Sort of busy with other things lately, so can't say how soon I can actually test it (since this means i'll also have to port other things like source locations to v3), but judging from your description this seems to be what I need.

vaartis:
Tested this one too and it seems to work fine :+1:

marzer:
Awesome, good to hear

PR #103, opened 2021-05-21
Added CMake FetchContent + tar archive downloading

proydakov:
Hi, marzer.

I added instruction for FetchContent + tar archive.
I think this might be useful for teams that use cmake and don't use git.

Could you review & accept my changes, pls ?


marzer:
Looks good, thanks :)

PR #114, opened 2021-09-21
build(meson): do not install if used as subproj

Tachi107:
**What does this change do?**

Prevents the headers and pkg-config file from being installed when the library is used as a subproject. 
Another possible approach would be to add a new option for disabling installation, but I don't see how installing a header-only library with the final executable would be useful (headers are only useful for development).

Nope, I directly created a PR instead of creating an issue and fixing it myself. 

**Pre-merge checklist**

- [ ] I've added new test cases to verify my change _(not necessary)_
- [ ] I've regenerated toml.hpp _(not necessary)_
- [ ] I've updated any affected documentation _(not necessary)_
- [ ] I've rebuilt and run the tests with at least one of: _(not necessary)_
    - [ ] Clang 6 or higher
    - [ ] GCC 7 or higher
- [ ] I've added my name to the list of contributors in [README.md](https://github.com/marzer/tomlplusplus/blob/master/README.md) (small change, I don't deserve it)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -496,10 +496,12 @@ if build_examples
 	subdir('examples')
 endif
 
-install_subdir('include/toml++',
-	strip_directory: true,
-	install_dir: 'include/toml++'
-)
+if not is_subproject
+	install_subdir('include/toml++',
+		strip_directory: true,
+		install_dir: 'include/toml++'
+	)
+endif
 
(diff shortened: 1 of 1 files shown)


marzer:
Ah, good catch! Thanks for your contribution.

PR #116, opened 2021-09-22
build(meson): use `override_dependency` if supported

Tachi107:
**What does this change do?**
This improves user experience when using toml++ as a Meson subproject.

With this change users will be able to simply call `dependency('tomlplusplus')` and have Meson automatically resolve the dependency from the subproject if it is not found on the system.

Without this, users always have to explicitly call `dependency('tomlplusplus', fallback: ['tomlplusplus', 'tomlplusplus_dep'])`

A lot of saved keystrokes :)

**Pre-merge checklist**

    - [ ] Clang 6 or higher
    - [ ] GCC 7 or higher

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -508,6 +508,10 @@ tomlplusplus_dep = declare_dependency(
 	version: meson.project_version(),
 )
 
+if meson.version().version_compare('>=0.54.0')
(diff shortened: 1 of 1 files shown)


marzer:
The hits keep on coming, thanks!

PR #141, opened 2022-02-01
build(meson): install CMake Config files to datadir

### LINKED ISSUE #140: build(meson): CMake's Package Version file is installed in the wrong place
## Environment
**toml++ version and/or commit hash:**  248e6031cfed41e24b3bdbed4da6780ff3e3978f to present

**Target arch:**  all

## Describe the bug

Quoting https://github.com/marzer/tomlplusplus/commit/248e6031cfed41e24b3bdbed4da6780ff3e3978f#commitcomment-65698096
>  1. The architecture-independent directory of CMake Config and Version files is `datadir/cmake`
> 2. CMake Version files are not architecture independent unless created with the `ARCHITECTURE_INDEPENDENT` option, and that is currently not supported by Meson (from the [CMake docs](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file) : _"an architecture check will be performed, and the package will be considered compatible only if the architecture matches exactly. For example, if the package is built for a 32-bit architecture, the package is only considered compatible if it is used on a 32-bit architecture, unless ARCH_INDEPENDENT is given"_)

I don't use personally use the CMake integration, but as a (newbie) Debian package maintainer I'm starting to care about architecture compatibilities. The most appropriate fix would be to set `ARCH_INDEPENDENT`, but that would require a patch in upstream Meson, so we currently have three options:

1. Keep using Meson's `write_basic_package_version_file()`, but move the output to the arch-dependent `libdir/cmake` - might require to also move the Config file, and it is not right
2. Manually write an arch-independent CMake Version file, and install that in `datadir/cmake` - I don't know if you ever looked at one, but they are not really human-friendly
3. Drop the CMake Version file and keep only the Config file, and wait for a fix in Meson (I'm working on it) - users would still be able to run `find_package(tomlplusplus)`, but without specifying the desired version

I haven't used Toml++ in quite some time, but I believe that this library can be consumed as both an header-only library and a compiled library, right? If so, neither the pkg-config file nor the CMake ones should unconditionally get installed in the architecture-independent directory, but that should depend on the current configuration (`libdir` when compiling the library, `datadir` when header-only)

Tachi107:
Since Meson doesn't yet support CMake's `ARCH_INDEPENDENT` option, a pre-generated Package Version file is installed instead of generating it at configure time.

I've also cleaned up a bit the nearby lines of code.

Fixes #140

**Pre-merge checklist**

    - [ ] Clang 6 or higher
    - [ ] GCC 7 or higher

### DIFF
--- a/cmake/tomlplusplus.cmake.in
+++ b/cmake/tomlplusplus.cmake.in
@@ -6,7 +6,7 @@ if(NOT TARGET tomlplusplus::tomlplusplus)
   # Import tomlplusplus interface library
   add_library(tomlplusplus::tomlplusplus INTERFACE IMPORTED)
   set_target_properties(tomlplusplus::tomlplusplus PROPERTIES
-    INTERFACE_INCLUDE_DIRECTORIES "${PACKAGE_PREFIX_DIR}/include")
+    INTERFACE_INCLUDE_DIRECTORIES "${PACKAGE_PREFIX_DIR}/@includedir@")
 
   # Require C++17
   target_compile_features(tomlplusplus::tomlplusplus INTERFACE cxx_std_17)
--- a/cmake/tomlplusplusConfigVersion.cmake.in
+++ b/cmake/tomlplusplusConfigVersion.cmake.in
@@ -1,0 +1,48 @@
+# This is a basic version file for the Config-mode of find_package().
+# It is used by write_basic_package_version_file() as input file for configure_file()
+# to create a version-file which can be installed along a config.cmake file.
+#
+# The created file sets PACKAGE_VERSION_EXACT if the current version string and
+# the requested version string are exactly the same and it sets
+# PACKAGE_VERSION_COMPATIBLE if the current version is >= requested version.
+# The variable CVF_VERSION must be set before calling configure_file().
+
+set(PACKAGE_VERSION "@version@")
+
+if (PACKAGE_FIND_VERSION_RANGE)
+  # Package version must be in the requested version range
+  if ((PACKAGE_FIND_VERSION_RANGE_MIN STREQUAL "INCLUDE" AND PACKAGE_VERSION VERSION_LESS PACKAGE_FIND_VERSION_MIN)
+      OR ((PACKAGE_FIND_VERSION_RANGE_MAX STREQUAL "INCLUDE" AND PACKAGE_VERSION VERSION_GREATER PACKAGE_FIND_VERSION_MAX)
+        OR (PACKAGE_FIND_VERSION_RANGE_MAX STREQUAL "EXCLUDE" AND PACKAGE_VERSION VERSION_GREATER_EQUAL PACKAGE_FIND_VERSION_MAX)))
+    set(PACKAGE_VERSION_COMPATIBLE FALSE)
+  else()
+    set(PACKAGE_VERSION_COMPATIBLE TRUE)
+  endif()
+else()
+  if(PACKAGE_VERSION VERSION_LESS PACKAGE_FIND_VERSION)
+    set(PACKAGE_VERSION_COMPATIBLE FALSE)
+  else()
+    set(PACKAGE_VERSION_COMPATIBLE TRUE)
+    if(PACKAGE_FIND_VERSION STREQUAL PACKAGE_VERSION)
+      set(PACKAGE_VERSION_EXACT TRUE)
+    endif()
+  endif()
+endif()
+
+
+# if the installed project requested no architecture check, don't perform the check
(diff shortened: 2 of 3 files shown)


Tachi107:
Uhm, this hardcodes the version number in yet another place. I can easily tell Meson to insert the version, so I'll do just that.
Converting to draft.

Tachi107:
@marzer since I'm working on the feature I described in https://github.com/marzer/tomlplusplus/issues/140#issuecomment-1025954587, could you please explain a bit to me how `TOML_HEADER_ONLY=0` and `TOML_IMPLEMENTATION` work? The docs don't go much into the details.

marzer:
The [Speeding up compilation](https://marzer.github.io/tomlplusplus/#mainpage-example-speed-up-compilation) section has what you need, I think

Tachi107:
I've read it, but it doesn't explain much.

Let's say I have a `toml.cpp` file with the following content

```cpp
#define TOML_HEADER_ONLY 0
#define TOML_IMPLEMENTATION
#include <toml++/toml.h>
```

And then I compile that single file as it was a regular library. In meson terms:

```meson
project('toml_lib', 'cpp')
lib = library('toml_lib', 'toml.cpp')
```

Can I then link against that library from external projects? Are shared symbols generated? How much of toml++ gets compiled in there, and how much of the implementation is kept into the headers?

marzer:
> but it doesn't explain much.

Well, given that it's a header-only library, it explains exactly what it needs to explain to support that use-case.

> How much of toml++ gets compiled in there, and how much of the implementation is kept into the headers?

The `.inl` files contain the implementation that is guarded behind `TOML_IMPLEMENTATION`. Give [toml.h](https://github.com/marzer/tomlplusplus/blob/master/include/toml%2B%2B/toml.h) a look to see how this works (it's a short file; just includes other files). The single-header `toml.hpp` is structured the the same as that file, just with the headers preprocessed inline.

What's not obvious from that file is _how much_ code is in the public `.h` files vs. the private `.inl` files, but the breakdown is approximately:
- Public code: ~7000 lines of mostly lightweight classes and templates
- Private code: ~9000 lines of actual code, of which approximately 4000 is the parser

> Let's say I have a toml.cpp file with the following content

In that particular example of just having the library included in a single cpp file, there is no difference between `TOML_HEADER_ONLY` being 0 or 1.

> Can I then link against that library from external projects? Are shared symbols generated?

That can be controlled by setting `TOML_API`. See [Library configuration](https://marzer.github.io/tomlplusplus/group__configuration.html)

Worth noting that is all pretty much the same as any other header-only library with an optional implementation target, like the stb libs, and a number of other popular ones. I'm not doing anything novel or tricky here.

Tachi107:
> > Are shared symbols generated?
> 
> That can be controlled by setting `TOML_API`. See [Library configuration](https://marzer.github.io/tomlplusplus/group__configuration.html).

Thanks, visibility support is really useful when working with ABI compatibility

> Worth noting that is all pretty much the same as any other header-only library with an optional implementation target, like the stb libs, and a number of other popular ones. I'm not doing anything novel or tricky here.

I've been in the coding world for just a few years (my first GitHub issue was opened in 2020), and I've never encountered header-only libraries that offer this kind of feature. So it might not be tricky to you, but is almost mind blowing to me :D

Now everything is super clear, I'll submit a PR when ready. Thanks!

marzer:
> So it might not be tricky to you, but is almost mind blowing to me :D

Hah, interesting. I almost wish it wasn't that common, to be honest; the only reason header-only libs like mine have become popular is as a workaround for the absolute clusterfuck that is the C++ ecosystem, lmao

marzer:
As a vaguely-related aside: here's a big list of libs with similar designs to this one: https://github.com/nothings/single_file_libs - lots of good stuff in there.

Tachi107:
> I almost wish it wasn't that common, to be honest; the only reason header-only libs like mine have become popular is as a workaround for the absolute clusterfuck that is the C++ ecosystem, lmao

Yeah that's true lol, I wonder <abbr title="ehm CMake ehm">[why](https://cmake.org)</abbr>. Anyway, could you please merge this PR so that I can branch of these changes to work on the other feature? I've also added a commit that avoids hardcoding `include` as the name of the include directory in the CMake Config file, so that it would work also when defining the `includedir` Meson option to a custom value

marzer:
> Anyway, could you please merge this PR so that I can branch off these changes to work on the other feature?

Sure thing. Sorry, I assumed you were just gonna bundle the other work into this one.

PR #142, opened 2022-02-02
wip: build(meson): implementing compile_library

Tachi107:
Still a lot of stuff to do.

https://github.com/marzer/tomlplusplus/issues/140#issuecomment-1025954587


marzer:
Do you want me to offer feedback as you work on the draft, or just wait until you mark it as ready?

Tachi107:
Feedback is always welcome, and now should be almost ready (I only need to define `TOML_API`)

meson.build
```diff
@@ -452,13 +452,13 @@ compiler_supports_int128 = compiler.links('''
...
-		#error __SIZEOF_FLOAT128__ wasn't defined!
+		#error __SIZEOF_FLOAT128__ wasn t defined!
 	#endif
 	#ifndef __FLT128_MANT_DIG__
-		#error __FLT128_MANT_DIG__ wasn't defined!
+		#error __FLT128_MANT_DIG__ wasn t defined!
 	#endif
 	#ifndef __LDBL_MANT_DIG__
-		#error __LDBL_MANT_DIG__ wasn't defined!
+		#error __LDBL_MANT_DIG__ wasn t defined!
```
Tachi107, line 461: Here I removed the `'` because they messed up my syntax highlighting, but I'll obviously put them back

Tachi107:
Changes so far:

- Added two meson.build files, one used when using toml++ as an header-only lib, one when using it as a compiled library. This is just for readability, because otherwise the main meson.build file would've got too large
- Renamed the CMake config files used by meson by appending `.meson` before `.in`, since the `cmake/` dir contains files shared by the two build systems, with similar names, and it is now easier to recognize who uses what
- created a `toml.cpp` file in a `lib` directory that is used to compile the implementation of the library
- Improved the `tomlplusplusConfig.cmake.meson.in` file to handle the compiled library

Tachi107:
Now failing due to a warning related to type attributes being in the wrong place

lib/toml++/toml.cpp
```diff
@@ -0,0 +1,9 @@
+#if !defined TOML_HEADER_ONLY || TOML_HEADER_ONLY == 1
+#	undef TOML_HEADER_ONLY
+#	define TOML_HEADER_ONLY 0
+#endif
+#ifndef TOML_IMPLEMENTATION
+#define TOML_IMPLEMENTATION
+#endif
```
marzer, line 7:
```suggestion

```

This isn't necessary; the lib implementation only has a single file - `toml.cpp` - so disabling `TOML_HEADER_ONLY` doesn't achieve anything.

Disabling `TOML_HEADER_ONLY` only makes a difference when including `toml.h` in multiple translation units.

include/toml++/impl/preprocessor.h
```diff
@@ -399,7 +399,27 @@
 
 // dllexport etc
 #ifndef TOML_API
-	#define TOML_API
+	#if TOML_HEADER_ONLY == 0
```
marzer, line 402:
```suggestion
	#if !TOML_HEADER_ONLY
```

Just to be sylistically consistent with the other preprocessor stuff in the library

include/toml++/impl/preprocessor.h
```diff
@@ -399,7 +399,27 @@
...
+		#if defined(_WIN32) || defined(__CYGWIN__)
+			#ifdef TOML_SHARED_LIB
+				#ifdef TOML_BUILDING_LIB
```
marzer, line 405:
```suggestion
			#if TOML_SHARED_LIB
				#if TOML_SHARED_LIB_BUILDING
```

I don't like just relying on things being defined; I prefer to have configurable `#defines` be treated as booleans wherever possible.

I'd also rename the 'building' one just so they both share the same prefix, making it very obvious that they're related.

lib/meson.build
```diff
@@ -0,0 +1,70 @@
+# Compiled library
```
marzer, line 1: I don't really have any comments about this file in particular, but I don't think `/lib` is the right place for the shared library implementation files; `/src` would be my preference. `/lib` to me suggests that I'd find static libraries or other build output in that directory.

marzer (changes requested):
Something else that occurs to me; is this also adding support for compiling as static library? If not, it probably should, since adding shared-lib support is going to open me up to feature requests along the lines of "shared, but no static option??? plz???"

Tachi107:
> Something else that occurs to me; is this also adding support for compiling as static library?

Yep

lib/toml++/toml.cpp
Tachi107: Oh ok, I thought it was mandatory since the docs state

> Enables the library's implementation when TOML_HEADER_ONLY is disabled.
> Not defined by default. Meaningless when TOML_HEADER_ONLY is enabled.

include/toml++/impl/preprocessor.h
Tachi107: I generally agree, but these are not options that should be exposed to users, so they shouldn't be already defined in any case. If you still think that checking their value is needed let me know

lib/meson.build
Tachi107: Ok, I chose it because I have a project that contains both library and an executable, so i put library headers in `include/`, library sources in `lib/` and sources of the executable in `src/`. But here it doesn't make much sense :)

include/toml++/impl/preprocessor.h
marzer: It's got nothing to do with _user_ configurability, I just prefer doing preprocessor-based configuration with `#if` over `#ifdef`.

meson.build
```diff
@@ -59,12 +59,7 @@ message('has_exceptions: @0@'.format(has_exceptions))
 
 # GCC or Clang
 if is_gcc or is_clang
-	add_project_arguments(
-		'-march=native',
-		'-fvisibility=hidden',
-		'-fvisibility-inlines-hidden',
-		language: 'cpp'
-	)
+	add_project_arguments('-march=native', language: 'cpp')
```
Tachi107, line 62: By the way, I think that silently passing `-march=native` to the compiler is not exactly ideal, as it could cause unexpected behaviour for users. It would also prevent every form of packaging, as the build machines' CPUs will always differ from users' ones.

marzer: Consider that previously the only compiled artefacts were test-related, and you'll understand why that flag is there. Now that you're adding a compiled library target it doesn't make sense, and should instead go in `tests/meson.build`.

lib/toml++/toml.cpp
marzer: That's the documentation for `TOML_IMPLEMENTATION`, not for `TOML_HEADER_ONLY`. Defining `TOML_IMPLEMENTATION` is only required _in one translation unit_ when `TOML_HEADER_ONLY == 0`, and is ignored altogether when `TOML_HEADER_ONLY == 1`.

Since this seems to be a recurring source of confusion:

| `TOML_HEADER_ONLY` | `TOML_IMPLEMENTATION` | What `toml.h` includes: |
|--------------------|-----------------------|-------------------------|
| 0                  | undefined             | Only the 'public' parts |
| 0                  | defined               | Everything              |
| 1                  | undefined*            | Everything              |
| 1                  | defined*              | Everything              |

*`TOML_IMPLEMENTATION` is ignored and does not need to be user-defined at all

include/toml++/impl/preprocessor.h
marzer: > Edit: also because they don't have a "default" value, their value depends on the current build configuration and can't be changed

```
#ifndef TOML_SHARED_LIB
				#define TOML_SHARED_LIB 0
#endif
```

```
-DTOML_SHARED_LIB=1
```

lib/toml++/toml.cpp
marzer: Also note that since `TOML_HEADER_ONLY == 1` is the default, any precompiled packaged version of the library will still need to have users manually define `TOML_HEADER_ONLY == 0` in their code/build system, otherwise this PR won't achieve anything except linker errors due to ODR violations.

Has that been addressed somewhere I haven't seen? I can't think of a way to solve that without changing `TOML_HEADER_ONLY` to be `0` by default, which is not a change I'm willing to make.

I guess you could programatically alter the default value for `TOML_HEADER_ONLY` during packaging by modifying `preprocessor.h`, but feels way too brittle.

Tachi107: Yes, I thought about that. The -DTOML_HEADER_ONLY=0 flags gets added to the pkg-config file, the CMake Config file and to the dependency object, so everyone consuming the compiled dependency will automatically get that flag. A bit like CMake's PUBLIC property

meson.build
Tachi107: Ok, I'll fix that too

lib/toml++/toml.cpp
marzer: Ah, I see. Sounds good to me.

meson.build
Tachi107: @marzer a lot of flags are currently added in the meson.build files, but none of them were previously used by the consumers of the library since nothing was actually compiled. Are they all intended to be used only for development? If so, leaving them in the `add_project_argiments()` call could create issues (like with `-march=native`). Can I move them all to an array named something like `develop_args` and apply them only in the `tests` directory?

Alright maybe no all of them, for example the flags that enable additional warning are not harmful in any way.

marzer: >  Are they all intended to be used only for development? 

Yup, that's more-or-less what I was saying above. Since previously there was no compiled code for the end-users, it greatly simplified my life to have them in the top-level project args.

> Can I move them all to an array named something like develop_args and apply them only in the tests directory?

I would like them to apply to the executable targets in `/examples` and `/toml-test`, too, but yes this sounds like a good solution.

Tachi107:
Now there are only two main issues to fix. The first one is the use of `TOML_API` in some templated classes, that GCC warns about:
https://github.com/marzer/tomlplusplus/blob/7b6101289fc210ae9c9fb50fb90dc963e1aeb770/include/toml%2B%2B/impl/value_extern.inl#L11-L24
GCC says: _"type attributes ignored after type is already defined"_

The other one is related to a message sent by `ci_single_header_check.py`

```log
Potentially missing #undefs:
	#undef TOML_SHARED_LIB
	#undef TOML_SHARED_LIB_BUILDING
```

Inspecting generate_single_header.py, I see that only "macros that are meant to stay public" should still be defined after the build.

`TOML_SHARED_LIB` is public; do I have to add it to that ignore_list?

`TOML_SHARED_LIB_BUILDING` is instead "private", but I would say that it less than private; it has to be defined only while building the shared library, and should never be set by the source files as they have no way of knowing if they are in the middle of the compilation of a shared library or not (kinda like `_WIN32`, it gets automatically defined based on the compiling environment). How should I handle it? Removing its definition from preprocessor.h would solve the issue, but it seems that you have different preferences so I don't know what should I do :)

marzer:
Yeah, for our purposes they're both 'public' w.r.t. that script. I'm happy for them to go in the ignore list.

meson.build
```diff
@@ -59,34 +59,28 @@ message('has_exceptions: @0@'.format(has_exceptions))
...
 
 # GCC
 if is_gcc
-	add_project_arguments(
+	devel_args += [
 		'-fmax-errors=5',
 		'-Wno-init-list-lifetime',
-		language: 'cpp'
-	)
+		'-DTOML_EXTERN_TEMPLATES=0'
```
marzer, line 80: why? it was never necessary here before. `TOML_EXTERN_TEMPLATES` is intended to solve a different problem.

marzer (changes requested):

Tachi107:
> It being set by a global header in the library itself is totally fine in the 'default' case. It's essentially an undocumented config option. I don't mind if it leaks.
> 
> ...having said that, you also make a reasonable point. Maybe `TOML_SHARED_LIB_BUILDING` should use `#ifdef` after all. I would still like `TOML_SHARED_LIB` to remain a boolean, though, as that better indicates the intent - "are you being used as a shared lib?" "yes/no"

I mean, for this simple case I believe that an `#ifdef` is just enough and doesn't complicate things, and doesn't expose a macro defined only because shared libraries work in an extraordinarily strange way on Windows... It's not only an implementation detail, it is an implementation detail of a particular platform/process that has little to do with this library in particular, and has no meaning outside of the compilation phase.

> Please remove `TOML_EXTERN_TEMPLATES` here. It was never necessary for the test+example applications before, so it shouldn't be now.

I don't really understand what that define does... But now _something_ is necessary to tell GCC to ignore the `TOML_API` macro when applied to `extern` variables. The issue was there before this PR, but `TOML_API` wasn't defined to anything when using GCC so nothing triggered it. If users were to define `TOML_API` themselves when using GCC they would've encountered the same issue (the one I described in https://github.com/marzer/tomlplusplus/pull/142#issuecomment-1029141355)

meson.build
```diff
@@ -77,7 +77,6 @@ if is_gcc
 	devel_args += [
 		'-fmax-errors=5',
 		'-Wno-init-list-lifetime',
-		'-DTOML_EXTERN_TEMPLATES=0'
```
Tachi107: Dropping EXTERN_TEMPLATES=0 did not helped :/

marzer: Yeah, I misunderstood what was happening before. What I didn't realize before, but do now, is that if you're now seeing `TOML_API`-related issues when building the 'devel' projects that you were not seeing before, then `TOML_API` is now being set to something it shouldn't be. They should not be consuming it as a shared/static library, but just directly via the header like before your changes, so `TOML_API` should not be set to anything.

> The issue was there before this PR,

Probably, for end-user code, sure. I don't doubt it. I've only used `TOML_API` on windows with DLL projects; I have not tested GCC in this scenario at all. Neither here nor there, though, because again, it never applied to the 'devel' projects, because `TOML_API` should not be set there. We can figure out what's going on for `TOML_API` on GCC once the test/CI projects are back to building how they were before.

marzer: Oh, I see what's changed. This addition is causing the issue:

https://github.com/marzer/tomlplusplus/blob/61b10632f388c1c46f2bf475a935caebdc326fde/include/toml%2B%2B/impl/preprocessor.h#L418-L422

You're unconditionally `__attribute__ ((visibility ("default")))` when compiling with GCC without header-only mode, even when not being consumed as a shared library. I think this is the wrong move. But I think I do understand the issue better now. Lemme think on this and I'll come up with something.

Tachi107: > You're unconditionally applying `__attribute__ ((visibility ("default")))` when compiling with GCC without header-only mode, even when not being consumed as a shared library.

Yep. I could set `TOML_API` to `__attribute__ ((visibility ("default")))` only when using the shared library, but that wouldn't change the fact that applying visibility attributes to `extern` variables is not possible with GCC. So the warning would still be there, but only when using the shared library. Compared to MSVC/Windows, using visibility attributes with GCC when using static libraries doesn't affect compilation/linking in particular ways

marzer: Ok, i know what needs to happen, but it's a bit much to explain here. I'll push some changes to the PR branch.

marzer: Decided I'm just gonna nuke the TOML_EXTERN_TEMPLATES nonsense since it makes this all much harder to reason about. We can fold it into your PR since it's basically a blocker for you.

Tachi107: Ok! Feel free to push your changes in this branch :)

marzer: Yup, just did, I removed the extern stuff altogether, which allowed me to simplify `preprocessor.h` quite a bit. Basically:

- `TOML_SHARED_LIB == 1` now implies/overrides `TOML_HEADER_ONLY == 0`
- I removed `TOML_SHARED_LIB_BUILDING` altogether, since it was serviing the exact same purpose `TOML_IMPLEMENTATION` was already
- `TOML_API` is only ever set to a non-blank value when `TOML_SHARED_LIB == 1`

That should simplify things enough to unblock you :)

include/toml++/impl/preprocessor.h
```diff
@@ -366,6 +366,11 @@
 #ifdef TOML_CONFIG_HEADER
 	#include TOML_CONFIG_HEADER
 #endif
+
+// is the library being built as a shared lib/dll using meson and friends?
```
marzer, line 370: Oh, eh. Guess this comment is technically wrong. I should have written:

```suggestion
// is the library being consumed as a shared lib/dll (using meson and friends)?
```

meson.build
Tachi107: Thanks! Overriding `TOML_HEADER_ONLY` makes sense. Regarding only setting `TOML_API` when building/consuming the shared library is I think that's also ok. Lastly, removing `TOML_SHARED_LIB_BUILDING` and only using `TOML_IMPLEMENTATION`... It took me a while to realize it, but you're right :D

include/toml++/impl/preprocessor.h
Tachi107: SHARED_LIB is defined during both compilation and usage, so I would not consider this as completely wrong

Tachi107:
We still have one issue left with symbol visibility: when linking against the compiled libraries, a symbol is missing. Here's the output of LLD:

```
ld.lld: error: undefined symbol: vtable for toml::v3::table
>>> referenced by main.cpp
>>>               CMakeFiles/linkami.dir/main.cpp.o:(toml::v3::table::table())
>>> referenced by main.cpp
>>>               CMakeFiles/linkami.dir/main.cpp.o:(toml::v3::table::~table())
>>> the vtable symbol may be undefined because the class is missing its key function (see https://lld.llvm.org/missingkeyfunction)

ld.lld: error: undefined symbol: vtable for toml::v3::node
>>> referenced by main.cpp
>>>               CMakeFiles/linkami.dir/main.cpp.o:(toml::v3::node::node())
>>> the vtable symbol may be undefined because the class is missing its key function (see https://lld.llvm.org/missingkeyfunction)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

And the one from gold:
```
CMakeFiles/linkami.dir/main.cpp.o:main.cpp:function toml::v3::node::node(): error: undefined reference to 'vtable for toml::v3::node'
/usr/bin/ld.gold: the vtable symbol may be undefined because the class is missing its key function
CMakeFiles/linkami.dir/main.cpp.o:main.cpp:function toml::v3::table::table(): error: undefined reference to 'vtable for toml::v3::table'
/usr/bin/ld.gold: the vtable symbol may be undefined because the class is missing its key function
CMakeFiles/linkami.dir/main.cpp.o:main.cpp:function toml::v3::table::~table(): error: undefined reference to 'vtable for toml::v3::table'
/usr/bin/ld.gold: the vtable symbol may be undefined because the class is missing its key function
collect2: error: ld returned 1 exit status
```

Tachi107:
Linking issues do not occur when setting the gnu_symbol_visibility to default (i.e. everything public), so this is almost surely because `TOML_API` is missing somewhere

marzer:
I think just unconditionally outlining the destructor is the 'standard' way to fix this, but I'm no longer at a PC to test. Happy to investigate tomorrow.

Tachi107:
Ok, I'll give it a shot myself. Have a good night!

marzer:
Fyi: circle ci didn't run your new job because you didn't add it to the workflow bit at the bottom of the scripr

marzer:
Relevant docs: https://circleci.com/docs/2.0/workflows/

marzer:
Sorry @Tachi107 I'm not going to get a chance to help with the linker thing today after all, won't be able to until Sunday or Monday at the earliest.

Tachi107:
Don't worry, take all the time you need

marzer:
@Tachi107 Had a chance to look at it just now. Turns out that yes, outlining some additional member functions did fix some of the linker issues, but another problem was the API annotations themselves needed to be a bit more granular than a one-size-fits-all macro. I've replaced `TOML_API` with `TOML_EXPORTED_CLASS` and `TOML_EXPORTED_FUNCTION` (the old one will still work if the two new ones aren't defined so existing users shouldn't be impacted).

Also needed to make some changes to the examples because they used a few inline library config options that caused linker failures with the compiled lib version, but that was pretty simple.

You can take it from here ^_^

Tachi107:
Awesome! I'll check that everything works a bit more before marking the PR as ready

marzer:
Ah, I probably should have removed the temp CI machinery you'd set up for github actions while I was there; looks like it's failing because it doesn't have locales initialized. CircleCI seems happy with the new compiled lib now though.

marzer:
Also, apologies, my API annotation changes make your PR look absolutely enormous now lmao

Tachi107:
I'll remove it later, as I also want to check that everything works on Windows (and I don't a pc with that OS here)

marzer:
You know what though, if github actions turns out to be alright I would consider migrating to it. Currently my CircleCI account is a paid one and I'd like to remove that expense from my life.

Tachi107:
I've been using GH Actions for a while and I can say that they are fine. Their default images are extremely bloated and can cause issues like conflicting libraries/dependencies, but you can always run your jobs in clean containers and they work really well

Tachi107:
When only building the DLL on Windows everything is fine, but when also building the tests it seems that Meson's "symbolextractor" fails... Are you able to reproduce this on your machine?

.github/workflows/compile.yaml
```diff
@@ -0,0 +1,75 @@
...
+        Remove-Item -Path C:\Strawberry -Recurse
+        choco install pkgconfiglite --yes --skip-virus-check --no-color
+        pip install meson ninja
+
+    - uses: ilammy/msvc-dev-cmd@v1
+
+    - name: Configure Meson
+      run: meson setup build --buildtype=debug -Dcompile_library=true -Dpedantic=true -Dbuild_tests=false -Dbuild_examples=true -Dgenerate_cmake_config=false -Db_lto=false
+      env:
+        CPPFLAGS: -DTOML_EXPORTED_CLASS=TOML_EXPORTED_FUNCTION
```
marzer, line 69: I don't think this will achieve anything; in the header file there's some logic that pre-emptively #undefs them both when `TOML_SHARED_LIB` is in use, and sets them manually

The rationale is that in windows DLL builds, you can `dllexport` a whole class, or `dllexport` specific functions _in_ a class, but not both

marzer:
> When only building the DLL on Windows everything is fine, but when also building the tests it seems that Meson's "symbolextractor" fails... Are you able to reproduce this on your machine?

Can't test it at the moment, but happy to give it a try later/tomorrow

.github/workflows/compile.yaml
Tachi107: Oh ok, I was just trying out random stuff 😅️

marzer:
Alright I've had some time to test it locally on my windows machine, and indeed using the `compile_library` option in meson was giving me all sorts of grief. Should be fixed now. The problem was two-fold:
- my new granular export settings weren't granular enough, and MSVC wasn't exporting the vtables properly
- the test script is basically a hard-coded compiler matrix, which you would normally do at the CI level, but since the lib has until now been exclusively header-only it was easy enough to just do it in meson. Meant the test code itself was written with this in mind. After some tweaks to the tests `meson.build` and simplifying some of the C++ test headers it compiled and ran OK

Also I realized that many of the options being put into `devel_args` (and thus not being passed onto the lib compilation were already covered by meson project-level options (e.g. `pedantic`) or were otherwise necessary to avoid overly-pendantic warnings and are otherwise harmless, so i've moved most of them back into a shared `universal_args` that is used by all the 'devel' projects _and_ the lib. Same goes for the top-level `overrides` array, which is now being correctly passed through to the lib.

marzer:
Hmmn, I note now that the windows job is still failing on the CI, but it's using clang, and hasn't really emitted any meaningful diagnostic output. I have no idea to debug that, though if a CI step is going to test windows, it should probably do so using MSVC, which is what I tested it locally with (and what most windows users will be using).

src/meson.build
```diff
@@ -1,23 +1,22 @@
...
 	include_directories: include_dir,
 	install: not is_subproject,
-	version: meson.project_version()
+	version: meson.project_version(),
+	override_options: overrides
 )
 
 tomlplusplus_dep = declare_dependency(
-	compile_args: args,
+	compile_args: lib_args,
```
Tachi107, line 19: Uhm... Maybe this is not ideal, as it would force _everybody_ to compile their projects with the compile flags used to build the library

Tachi107: I mean, not all flags are meant to be public, only the ones required to consume the library (i.e. things like `-DTOML_SHARED_LIB=1`). Exposing everything would not only be technically wrong, but could also cause compile failures for users (for example, if you compile the lib with Clang and a user consumes the pkg-config file and then compiles their project with GCC, the compilation would fail because it wouldn't be able to understand flags like `-Weverything`).

This is why I left them in `devel_args`, because while they are not _required_ to build the library, they are mostly interesting for you, the developer, so that you can take a look at all the warnings and fix them. But users shouldn't really care

Tachi107: Example:

```text
$ CXX=clang++ meson setup build --prefix="$(pwd)/install" -Dcompile_library=true
$ meson install -C build
$ cat install/lib/x86_64-linux-gnu/pkgconfig/tomlplusplus.pc
prefix=/home/tachi/Scrivania/tomlplusplus/install
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: tomlplusplus
Description: TOML config file parser and serializer for C++
URL: https://marzer.github.io/tomlplusplus
Version: 3.0.1
Libs: -L${libdir} -ltomlplusplus
Cflags: -I${includedir} -ferror-limit=5 -Wno-unused-command-line-argument -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-documentation -Wno-documentation-unknown-command -Wno-switch-enum -Wno-covered-switch-default -DTOML_HEADER_ONLY=0 -DTOML_SHARED_LIB=1

```

marzer: Oh, right. I guess I've misunderstood the relationship between the dependency flags and flags for just one target. I didn't realise it was passing the flags onto users consuming the library too.

Is there some way to apply the 'devel' flags to the lib when it is built, but not pass them on to the users? I'd still like the optional pedantic warning setting to apply if I choose it.

Tachi107:
> Alright I've had some time to test it locally on my windows machine, and indeed using the `compile_library` option in meson was giving me all sorts of grief. Should be fixed now. The problem was two-fold:
> 
>   * my new granular export settings weren't granular enough, and MSVC wasn't exporting the vtables properly
> 
>   * the test script is basically a hard-coded compiler matrix, which you would normally do at the CI level, but since the lib has until now been exclusively header-only it was easy enough to just do it in meson. Meant the test code itself was written with this in mind. After some tweaks to the tests `meson.build` and simplifying some of the C++ test headers it compiled and ran OK

Nice :)

> Also I realized that many of the options being put into `devel_args` (and thus not being passed onto the lib compilation) were already covered by meson project-level options (e.g. `pedantic`) or otherwise harmless (disabled warnings etc.), so i've moved most of them back into a shared `universal_args` that is used by all the 'devel' projects _and_ the lib. Same goes for the top-level `overrides` array, which is now being correctly passed through to the lib.

See my previous comment (https://github.com/marzer/tomlplusplus/pull/142#discussion_r800189394), but I'd also like to give my personal opinion here. Playing around with flags is nice when developing stuff, but I've found out that when you are compiling a binary that has to be shipped onto user's machines it is best to leave things as close as the defaults as possible. Because some flags are maybe not so well tested, or because you can never know what flags will be applied downstream (for example, Debian passes a lot of hardening flags to the compiler) and conflicts may arise.

> Hmmn, I note now that the windows job is [still failing](https://github.com/Tachi107/tomlplusplus/runs/5083533066?check_suite_focus=true) on the github actions CI, but it's using clang, and hasn't really emitted any meaningful diagnostic output. I have no idea how to debug that, though if a CI step is going to test windows, it should probably do so using MSVC, which is what I tested it locally with (and what most windows users will be using). I don't think explicitly supporting building as a DLL using clang on windows is a worthy venture.

See, that's the bloat I was talking about... GH images are full of every single peace of software you could ever think of (GCC, Clang, MSVC, JavaScript stuff, python, perl, package managers...), and while this can be handy at times (especially on Windows where you don't have package managers like apt) it almost surely causes issues. I could look into using a container also for the Windows job.

> (not to say I don't want people to use that specific combo, but that it is more niche and not really worth the support burden)

I'm not sure how niche it is really, as Clang outputs really performant code compared to MSVC. Chrome is compiled with Clang on Windows, for example.

> Edit: Same goes for cygwin, really. If someone hates themselves enough to still be using that dumpster-fire when WSL is _right there_, then the header-only version should suffice smiling_face_with_tear

Yeah I don't see why you would want to use that over WSL

marzer:
> See my previous comment (https://github.com/marzer/tomlplusplus/pull/142#discussion_r800189394), but I'd also like to give my personal opinion here. Playing around with flags is nice when developing stuff, but I've found out that when you are compiling a binary that has to be shipped onto user's machines it is best to leave things as close as the defaults as possible.

Well yes, I get that. I never intended to pass that stuff on to users. I want the extra 'devel' flags etc. to be available to _me_ when building the lib, but not on by default (instead of just unconditionally compiling the lib with default flags and completely ignoring my preferences, which is what you were doing). I have a few options for development/testing purposes in `meson_options.txt`, so the compiled version of the lib should respect them just as the tests etc. do already.

> I'm not sure how niche it is really, as Clang outputs really performant code compared to MSVC. Chrome is compiled with Clang on Windows, for example.

That is a poor example because that is in fact a very niche build configuration. Google-tech is notoriously 'NIH' when it comes to their toolchain. It's not at all representative of the general case. Besides, I have zero personal desire to offer support for that use-case if/when it breaks - unless you're gonna stick around as maintainer and debugger for it basically forever, it's not useful to lock-in now. As said, people with those needs can use the library in header-only mode (which, in terms of build flexibility, is strictly superior to anything we'll ever do here in this PR).

Tachi107:
> I never intended to pass that stuff on to users. I want the extra 'devel' flags etc. to be available to _me_ when testing the compiled version of the lib, but not on by default (instead of just unconditionally compiling the lib with default flags and completely ignoring my preferences, which is what you were doing, and what I attempted to correct). I have a few options for development/testing purposes in `meson_options.txt`, so the compiled version of the lib should respect them just as the tests etc. do already.

Ok, we could:

1. Add an option to enable those flags
2. Only pass those flags to the compiled library without making them public (giving this a try right now)


> That is a poor example because that is in fact a very niche, very highly-specialized build configuration. Google-tech is notoriously 'NIH' when it comes to their toolchain. It's not at all representative of the general case.

Yeah that's a bad example, and I have no doubts that MSVC is more popular thanks to its historical integration with Visual Studio. But I seem to remember that the LLVM toolchain has been integrated in Visual Studio as well, and that helps with adoption; at least I used it when I was on Windows.

In the end I honestly don't care about it, I was mostly interested in this PR because I wanted to package this great library for Debian as there are not many high-quality alternatives there, but I'm really curious to know why it fails, so I'll open a Windows VM and test stuff there :)

One last thing. I'll have an important exam in a few days, so I'll probably stop working on this until the end of next week. Sorry!

meson.build
```diff
@@ -211,17 +195,18 @@ endif
...
 	overrides += 'strip=true'
 endif
 
+# library is being compiled as a separate lib target
+if get_option('compile_library')
+	universal_args += '-DTOML_HEADER_ONLY=0'
+	if get_option('default_library') != 'static'
+		universal_args += '-DTOML_SHARED_LIB=1'
+	endif
+endif
```
Tachi107, line 209: These shouldn't be needed, because you only need to use them if you're using the compiled library. So they should be set as `compile_args` of the `tomlplusplus_dep` object, so that every consumer (in our case the tests and examples) uses them

Tachi107:
Why are only 12 tests ran when `compile_library` is true?

marzer:
Because this:
> the test script is basically a hard-coded compiler matrix, which you would normally do at the CI level, but since the lib has until now been exclusively header-only it was easy enough to just do it in meson. Meant the test code itself was written with this in mind. After some tweaks to the tests meson.build and simplifying some of the C++ test headers it compiled and ran OK

For the compiled library it turns off the permutations based on `-fast-math`, `-std=c++XX`, etc., and just uses the 'default' so the tests/examples match the compiled lib

Have a look inside `tests/meson.build` if you want to see where this happens. There's a few nested loops to handle the different permutations.

Tachi107:
Investigated a bit during a break, and I found out that the Clang Windows build fails because of LTO. I guess Windows' `link.exe` can't handle Clang's LTO format

include/toml++/impl/preprocessor.h
eli-schwartz: The difference between `#if` and `#ifdef` can bite you really hard even as the developer of a project, for example when porting between build systems. Accidentally forget to port one check over:
- use of `#ifdef` causes the compiler to effectively think the check returned false
- `#if` errors out since you didn't define it one way or another, and forces you to go back and check whether you actually did the check and defined the result.

eli-schwartz:
> When only building the DLL on Windows everything is fine, but when also building the tests it seems that Meson's "symbolextractor" fails... Are you able to reproduce this on your machine?

This is a pip bug: https://github.com/mesonbuild/meson/issues/9955
Ultimately the fault of https://github.com/pypa/pip/issues/10875

Downgrade pip and meson will have a valid `sys.stderr` again.

Tachi107:
Thanks, I had _no idea_ what was going on :)

marzer:
>i'm loosing all hope
> but maybe something now works
> or maybe not

> something is working

good old CI lmao

eli-schwartz:
The only thing worse than fighting with CI is fighting with production systems because you didn't have CI...

Tachi107:
@marzer, I managed to get the Windows container CI working :D

It's a mess, and I don't know it using it is really worth it compared to the bloated GitHub Actions images, but it is there if you want

Tachi107:
I think this is almost ready. If you're ok with the changes I could edit the commit history and keep only a few meaningful commits (like "depreate TOML_API", "add compile_library option", "switch to GitHub Actions", etc).

Normally I would squash everything in one single commit, but since this PR is so huge I believe that having four or five different commits in the master branch would be better.

marzer:
Yeah, sure, that sounds like a good idea.

Tachi107:
@marzer mostly done (I'm pinging you just in case you're not subscribed to this thread). 

I only need to move the `generate_dox` and `deploy_dox` jobs to GitHub Actions, but I don't fully understand how they work, or better, what they do.

If what they need to do is to build the html files and deploy them to GitHub Pages then I could add a job similar to the one I added in [pistacheio/pistache/.github/workflows/pistache.io.yaml](https://github.com/pistacheio/pistache/blob/master/.github/workflows/pistache.io.yaml)

.clang-format
```diff
@@ -165,6 +165,9 @@ StatementMacros:
   - TOML_ATTR
   - TOML_CONST_GETTER
   - TOML_CONST_INLINE_GETTER
+  - TOML_EXPORTED_CLASS
+  - TOML_EXPORTED_MEMBER_FUNCTION
+  - TOML_EXPORTED_FREE_FUNCTION
```
marzer, line 170: Ah, looks like I forgot `TOML_EXPORTED_STATIC_FUNCTION` here. Mind adding it for me?

.github/workflows/gh-pages.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Install dependencies
+        run: |
+          apt-get -y update
+          apt-get -y install --no-install-recommends python3-pip
+          pip3 install --upgrade --requirement tools/requirements.txt
+
+      - name: Generate docs
+        run: |
+          cd docs
+          poxy --verbose
```
marzer, line 30: [`poxy`](https://github.com/marzer/poxy) requires a relatively recent version of Doxygen (the docker image I was using had v1.9.2). This image might need it installed or updated for poxy to work correctly.

README.md
```diff
@@ -145,14 +145,16 @@ won't need to mess with these at all, but if you do, set them before including t
...
 | `TOML_ENABLE_UNRELEASED_FEATURES` |     boolean    | Enables support for [unreleased TOML language features].                                                 | `0`                    |
 | `TOML_ENABLE_WINDOWS_COMPAT`      |     boolean    | Enables support for transparent conversion between wide and narrow strings.                              | `1` on Windows         |
 | `TOML_EXCEPTIONS`                 |     boolean    | Sets whether the library uses exceptions.                                                                | per compiler settings  |
+| `TOML_EXPORTED_CLASS`             |     define     | API export annotation to add to classes.                                                                 | undefined              |
+| `TOML_EXPORTED_MEMBER_FUNCTION`   |     define     | API export annotation to add to non-static class member functions.                                       | undefined              |
+| `TOML_EXPORTED_FREE_FUNCTION`     |     define     | API export annotation to add to free functions.                                                          | undefined              |
```
marzer, line 157: Eh. I forgot `TOML_EXPORTED_STATIC_FUNCTION` here too 😅

marzer (changes requested):

.github/workflows/gh-pages.yaml
Tachi107: Is the [`doxygen`](https://packages.debian.org/testing/doxygen) package enough or does it need extra components like [`graphviz`](https://packages.debian.org/testing/graphviz) (containing the `dot` executable)?

marzer: Doxygen alone should suffice :)

Tachi107:
I've also updated the badge in the Readme to point to the GitHub CI job, and now I really think that this is complete. When merging, be careful not to squash the commits :)

marzer:
Thanks for all your work on this!

PR #143, opened 2022-02-12
build(meson): use system deps when available

Tachi107:
This is especially important in cases where you want to run tests but you have no internet access, nor Git submodules, like when building Linux packages.

I had to move the submodules from `external` to `subprojects`, as Meson uses that folder when looking for fallbacks (more info [here](https://mesonbuild.com/Subprojects.html#why-must-all-subprojects-be-inside-a-single-directory)).

I've also added an extra boolean in tests/meson.build that avoids building the tl_optional tests when the `tloptional` submodule is unavailable.


meson.build
```diff
@@ -42,7 +42,7 @@ is_windows = host_machine.system() == 'windows'
 is_x64 = host_machine.cpu_family() == 'x86_64'
 is_subproject = meson.is_subproject()
 has_exceptions = get_option('cpp_eh') != 'none'
-include_dirs = include_directories('include', 'external')
+include_dirs = include_directories('include', 'subprojects')
```
eli-schwartz, line 45: Please exclusively use `dependency()`, meson gets mad at you if you try to reference `subprojects/` as a path.

Due to an implementation bug, it didn't warn you if you literally used `subprojects/`, which I'm fixing in https://github.com/mesonbuild/meson/pull/9997

tests/meson.build
```diff
@@ -26,6 +26,24 @@ test_sources = [
 	'windows_compat.cpp'
 ]
 
+test_deps = []
+
+test_deps += dependency('nlohmann_json', fallback: ['json', 'nlohmann_json_dep'])
```
eli-schwartz, line 31: You should just use the WrapDB wraps, which include the `fallback` information in Meson 0.54+ compatible format encoded in the wrap file, and this project needs that Meson version anyway.

tests/meson.build
```diff
@@ -26,6 +26,24 @@ test_sources = [
...
+if catch2_dep.found()
+	test_deps += catch2_dep
+else
+	run_command('git', 'submodule', 'update', '--init', '--depth', '1', meson.source_root()/'subprojects'/'Catch2')
+endif
+
+# When building offline, skip the tloptional test
+has_tloptional = false
+run_command('git', 'submodule', 'update', '--init', '--depth', '1', meson.source_root()/'subprojects'/'tloptional', check: false)
+if import('fs').exists(meson.source_root()/'subprojects'/'optional'/'include'/'tl'/'optional.hpp')
```
eli-schwartz, line 43: If only this were a wrap, and you could just check if the subproject is available. ;)

eli-schwartz (changes requested):

tests/meson.build
marzer: > I didn't want to diverge too much from the existing approach

Nah. I'm a build system novice. Probably lots of bad in there. If you have ways to clean it up, I encourage it - I'll probably learn from it.

tests/meson.build
```diff
@@ -26,6 +26,24 @@ test_sources = [
...
+test_deps = []
+
+test_deps += dependency('nlohmann_json', fallback: ['json', 'nlohmann_json_dep'])
+
+catch2_dep = dependency('catch2', required: false)
+if catch2_dep.found()
+	test_deps += catch2_dep
+else
+	run_command('git', 'submodule', 'update', '--init', '--depth', '1', meson.source_root()/'subprojects'/'Catch2')
+endif
```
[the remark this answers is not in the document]
marzer: This is a leftover from my original hackery.

tests/meson.build
Tachi107: Using [wraps](https://mesonbuild.com/Wrap-dependency-system-manual.html) would mean that the test dependencies will be available only when using Meson. This has the advantage of cleaning things up considerably, but on the other hand you would be able to run tests only when building the project with Meson. You would still be able to use Visual Studio to develop the library as Meson has a VS backend that generates sln files, but I don't want to impose you to change your development workflow :)

Tachi107: I've updated `CONTRIBUTING.md` to include instructions about how to test with Meson on Windows and Linux. Note that I'm not a native English speaker, and the grammar there might not be perfect

eli-schwartz: Nice! Looks so much cleaner. :D

marzer: > Using [wraps](https://mesonbuild.com/Wrap-dependency-system-manual.html) would mean that the test dependencies will be available only when using Meson.
> You would still be able to use Visual Studio to develop the library as Meson has a VS backend that generates sln file

Yeah, I'd like to avoid that. I am very comfortable with MSBuild and the VS ecosystem (in fact it's about the only one I'd say I'm 'good' at, so it's basically my native state.

Can I still clone git submodules with this method, or does that replace submodules altogether?

marzer: To be clear: I don't mind if the dependencies require meson on linux (or windows, if someone wants to go that route), but I don't want to have do anything differently as the maintainer of the library.

marzer: For the sake of completeness, my workflow for this project is thus:
- Dev using VS on windows
- Test directly in VS, allowing me to use the VS debugger natively rather than woeful gdb & lldb
- Run tests directly in VS via the Catch2 test adapter extension
- Hooray, I've done most of the work all within the same interface without having to touch a command-line at all
- Invoke meson to test with clang + gcc on linux using WSL Ubuntu - sometimes this yields fun surprises but mostly just compiler warnings for things not covered by MSVC

Also worth nothing, the Ubuntu WSL part is from the same physical repo folder on my drive (i.e. it operates out of the mapped path of the windows folder in the linux file system, so like `/mnt/d/Repositories/tomlplusplus` or something), which is why the submodules approach was a good one. I started with a maually-curated VS solution and 'patched in' the linux stuff on top. Since I'm very comfortable in VS land, this works well for me - meson is a means to an end.

eli-schwartz: Meson doesn't care whether you use tarballs like this PR, or git submodules. In fact, meson wraps can use several VCSes including git, tarballs are just one option (which is zero-dependency and thus recommended, not everyone builds from git checkouts of tomlplusplus).

However, Meson does require that however you get the subproject, it needs a `meson.build`. The tarballs provide this, because the .wrap files being checked in here are distributed by a Meson infrastructure that sets all this up for you. Specifically it applies "meson build overlays" on top of upstream release tarballs for non-native projects.

So, there is a snag. As it happens, nlohmann_json is already a meson-native project, you may freely use git submodules there. (I personally wouldn't, the upstream repository is huge...)

The others aren't, and need a zip archive built for the release tarballs.

So, I can see three options here:
- support MSVC by running `meson subprojects download` rather than `git submodule update --init`, this requires meson as a build dependency but doesn't force you to use it as the build ***driver***
- include meson wraps, but for MSVC include a second copy of the dependencies as submodules in idk, `msvc/external/`
- import the Meson WrapDB version of the dependencies by checking the `subprojects/packagefiles/foo/meson.build` into git. Write your own wrap files for each one, which are inspired by the Meson WrapDB but not installed from it. These will use the `[wrap-git]` style, list the clone url, and include a `patch_directory =` rather than patch_url, which merges local overlays from `subprojects/packagefiles/`. The end result is that meson can clone the submodule for you, even when tomlplusplus is not cloned from git it can use the clone url duplicated in the wrap file. But ultimately it's a git submodule and does  everything you'd expect out of that, such as be usable with bare git and MSVC.


The minor downside of option 3 is that you end up checking a bit more lines of code into git, but it's really not that much. My gut feeling is that it will be your favorite option.

eli-schwartz: Oh, having seen your last comment I guess option 1 might be fine with you too.

Note that if you use `git submodule` to download dependencies that don't have native meson.build, meson will get confused, think the subproject is already there, and not apply the overlay. Then later another part of meson fails because meson.build is missing. (This is a bit unfortunate, but on the other hand meson really doesn't want to accidentally overwrite anything you've manually done in there. Some people use wrap filea to hack on dependencies and then push to the dependency's remote.)

You can manually apply the overlay with `meson subprojects packagefiles --apply` but also you can just use meson's download command in the first place.

marzer: I mean... there is an implicit fourth option: do none of that and just leave it how it was. I don't really care about meson. As far as I'm concerned it just builds and runs tests on linux. The library is primarily a header-only one precisely _because_ I didn't want to have to care about linux package managers etc.

eli-schwartz: Linux package managers will typically never use this. :D They prefer `apt install catch2` instead of submodules.

Meson wraps are primarily a *dev* package manager, because people who add tomlplusplus to their project with `meson wrap install tomlplusplus` don't have to fiddle with package managers at all, and meson will download json, catch2, etc. on demand.

Wraps also do built-in delegation to a version installed through `apt`, for the people that prefer that.

Anyway options 2 and 3 are both fairly good in terms of being "set up once and forget".

eli-schwartz: > So long as all it does is fetches the dependencies and sticks them in a stable place locally, such that my VS workflow still works untouched by meson.

That is indeed what it does. You just end up with slightly different paths to where the dependencies are located (and one new meson-specific file copied in).

marzer: Well in that vein, people who add tomlplusplus to their project won't need any of the dependencies anyway. All of toml++'s dependencies are used _only_ in testing. They are not at all required for regular consumption of the library - that only requires the C++ standard library.

marzer: But also, returning to the root goal of this PR: removing git submodules because apparently not everyone will have access to git or internet connection etc.

The more I think about it, the less understand what value is being added here. Since the dependencies are all for testing, and testing is for development, why is git and internet access even worth thinking about? Someone without those things can't work on the library anyway.

eli-schwartz: People often do want to run tests on their direct dependencies too, and meson allows toggling this (project options to set up tests, test options to select the main suite or the tomlplusplus suite...).

For example they may be building on an OS or CPU architecture or compiler you haven't got in your CI.

Tachi107: > include meson wraps, but for MSVC include a second copy of the dependencies as submodules in idk, msvc/external/

I think that this is the more viable. @marzer doesn't have to change his workflow and users can run tests in restricted environments (i.e. offline).

> there is an implicit fourth option: do none of that and just leave it how it was. I don't really care about meson.

We could, but then toml++ would offer a worse experience for a lot of users (Linux distributions, for example, need to be able to build the library and run tests offline and without git submodules/wraps, only using installed libraries as dependencies).

Of course you can always try and learn new stuff, for example you could try using Meson on windows for a while and see how that works out for you (at first you could also generate VS solutions with `meson setup --backend=vs2022`). When you're not tied to Visual Studio you gain a lot of flexibility, you could develop your stuff on other IDEs, on code editors like VSCode (+ clangd), you would even be free to write code on other OSses without changing your workflow :D (or at least that's what I've done, and I'm really happy now. Sorry for this little digression).

marzer: > Of course you can always try and learn new stuff, for example you could try using Meson on windows for a while and see how that works out for you (at first you could also generate VS solutions with meson setup --backend=vs2022).  When you're not tied to Visual Studio you gain a lot of flexibility, you could develop your stuff on other IDEs, on code editors like VSCode (+ clangd)

I have worked like that in the past with other projects/at work, with CMake, and still prefer VS. I'm not _tied_ to VS, I explicitly _choose_ VS.

> you would even be free to write code on other OSses without changing your workflow

I'm hitting windows and linux with this workflow already. That's enough for me.

marzer: > include meson wraps, but for MSVC include a second copy of the dependencies as submodules in idk, msvc/external/

If we're going to go down that route, the previous directory of `/external` would be my choice. The structure of the repo is based on the `pitchfork` layout.

Tachi107:
Tests are now failing on Windows + vs2022 with `#error TOML_COMPILER_EXCEPTIONS was not deduced correctly`

tests/lib_catch2.h
```diff
@@ -24,6 +24,8 @@
 
 #if __has_include(<Catch2/single_include/catch2/catch.hpp>)
 #include <Catch2/single_include/catch2/catch.hpp>
```
Tachi107, line 26: Including headers this way looks like a hack. Isn't it possible to tell Visual Studio to add the various subprojects to the include path for tests?

marzer: It already is set up to do that; the top-level Catch2 is the subproject.

I could have easily added the deeper path, of course. Doing it this way was mostly out expediency - there was never any danger of it being incorrect as it was a git submodule tied to a specific commit.

tests/meson.build
marzer: Just saw this:

> We could, but then toml++ would offer a worse experience for a lot of users (Linux distributions, for example, need to be able to build the library and run tests offline and without git submodules/wraps, only using installed libraries as dependencies).

Please, _please_ understand: the library is header-only. The idea that anybody _needs_ to build the library is incorrect. It is 'compilable' only because you wanted to add that, but I 100% do not care about that at all, and never will. I entertained the idea because it seemed somewhat painless for me, though that appears to be changing. I do not care about linux distributions. I do not care about package managers. I do not care about build systems. I do not care about situations where people can't use git. Whatever, use a different library. There's going to be very few people for whom that will actually matter in practice. I wrote toml++ as header-only precisely because I didn't want to have to care about that. It is literally a single piece of C++ code you drop directly into your project. I do not want to change that. If it looks like your recent changes are inviting scope creep of the sort of "package manager this", "linux distros that", I will simply revert them, rather than invite further scope creep.

To repeat: Nobody, living or dead, past or present, on this level of reality or somewhere on a higher plane, needs to explicitly build toml++. It is header-only. Please make no further appeals contrary to this.

Tachi107: > I don't see how your reply is related. I was not suggesting the subprojects directory be renamed. I was saying that it should be simply `external/` instead of `msvc/external/` if we were going to have two separate copies of the depedendencies according to workflow. That approach is broken for other reasons, of course: it has two independent copies of the dependencies, which is goofy as hell

Oh ok, I misinterpreted it. And yes having both a submodule and a wrap is inconvenient. If you prefer to only keep submodules let me know. The meson.build files will be less clean, but maybe that's fine for you.

> Nobody, living or dead, past or present, on this level of reality or somewhere on a higher plane, needs to explicitly build toml++. It is header-only first and foremost. Please make no further appeals contrary to this.

I often mention compiling the library because that's what I would like to do, but these changes are not needed for that (everything was already done in my other PR). These changes are needed to use system dependencies for tests when available. All the other things I talked about are situations where using system dependencies is beneficial / the only thing that is possible to do. For example, cloning the nlohmann-json submodule is _painful_ for people like me who have a slow network connection (~3Mb/s).

marzer: > Oh ok, I misinterpreted it. And yes having both a submodule and a wrap is inconvenient. If you prefer to only keep submodules let me know. The meson.build files will be less clean, but maybe that's fine for you.

Well of course I'd prefer that! That's basically what I've been saying the whole time. That fits my current workflow perfectly, because _it is_ my current workflow. The meson.build file being messy doesn't bother me in the slightest (I mean, you've seen my meson.build files already, so you know that :P). Somehow over the course of this discussion we ended up replacing submodules with wraps, forcing me to use meson on windows, etc... seems like gradually this moved further and further away from the original intent, and I was (and am still somewhat) concerned about the potential for "ooh shiny" build-system stuff. Might have been my mistake initially for giving the work my blessing, but I didn't realize what I was encouraging. I just expected some minor `meson.build` refactoring.

> For example, cloning the nlohmann-json submodule is painful for people like me who have a slow network connection (~3Mb/s).

Well that's fair. Still, if you're just a regular user of the library you don't need the tests, You might _want_ them, sure, but I don't consider a bit of tedium in that situation a barrier, since you're already opting-in to something you didn't previously need to worry about. Hence me not wanting to inconvenience myself for the sake of some eager beavers. If you can make the system deps work without it being a royal PITA _for me specifically_, then fantastic, I'm on board.

(as an aside: the "cloning nlomann/json submodule sucks" problem can generally be avoided using `--depth 1`)

Tachi107: Oh and also note that I like to mess with stuff. That is, I proposed/pushed stuff to hear your feedback and also to see how much everything I did broke CI jobs, of course I don't want you to take in everything I propose. I'm just a kid that is still learning how to write code, after all :)
If we also count the language barrier we have the perfect recipe for misinterpretations and (constructive) arguments :D

tests/meson.build
```diff
@@ -26,17 +26,29 @@ test_sources = [
...
-]
+test_deps = [tomlplusplus_dep]
+
+fs = import('fs')
+
+catch2_dep = dependency('catch2', required: false)
+if catch2_dep.found()
+	test_deps += catch2_dep
+elif fs.exists('..'/'external'/'Catch2'/'single_include')
+	catch2_dep = declare_dependency(include_directories: '..'/'external'/'Catch2'/'single_include')
```
Tachi107, line 37: Yeah this breaks Meson's sandbox, but nothing is perfect

tests/meson.build
eli-schwartz: Right, depth can help here because of https://github.com/nlohmann/json/commit/dbf1a1f413075f5c57bdc1c6f1df0398dd440686

Since 3.8.0 the nlohmann_json repository no longer includes test data directly, but downloads it from https://github.com/nlohmann/json_test_data.git which you can avoid doing if you don't need it.

toml-test/meson.build
```diff
@@ -1,9 +1,14 @@
+tt_deps = [
+	tomlplusplus_dep,
+	dependency('nlohmann_json', fallback: ['json', 'nlohmann_json_dep'])
```
Tachi107, line 3: This does not require the same `declare_dependency()` hacks that I used in tests/meson.build because the `json` submodule contains a `meson.build` file

meson.build
```diff
@@ -4,6 +4,7 @@ project(
 	version: '3.0.1',
 	meson_version: '>=0.54.0',
 	license: 'MIT',
+	subproject_dir: 'external',
```
marzer, line 7: Was this necessary, in the end? Is this what makes `catch2_dep = dependency(...)` etc work?

PR #156, opened 2022-05-25
Path

### LINKED ISSUE #153: Update value at path
**Is your feature request related to a problem? Please describe.**
It would be very useful to be able to insert or modify a value at a particular path, essentially the inverse of the at_path(...).value_or(...) method chain.

**Describe the solution you'd like**
Something like tbl.at_path(...).insert_or_update(...). At the moment it seems like I would need to parse the path string myself to find the correct parent table to then call insert_or_assign, which is unfortunate since the path-parsing code already exists in the library. Or is there a canonical way of doing this already that I'm missing?

jonestristand:
**What does this change do?**
This PR adds an implementation of toml::path which represents a path to an arbitrary toml node. It provides utilities for path manipulation by path segment (e.g. parents, truncation), as well as comparisons between paths. toml::path objects may be used as an argument for at_path.

Fixes #153.

**Pre-merge checklist**

- [x] I've updated any affected documentation (comments updated, have not regenerated with poxy)
    - [ ] Clang 6 or higher
    - [ ] GCC 7 or higher
    - [x] MSVC 19.20 (Visual Studio 2019) or higher


include/toml++/impl/std_algorithm.h
```diff
@@ -0,0 +1,12 @@
+//# This file is a part of toml++ and is subject to the the terms of the MIT license.
+//# Copyright (c) Mark Gillard <mark.gillard@outlook.com.au>
+//# See https://github.com/marzer/tomlplusplus/blob/master/LICENSE for the full license text.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+#include "preprocessor.h"
+TOML_DISABLE_WARNINGS;
+#if !TOML_HAS_CUSTOM_OPTIONAL_TYPE
+#include <algorithm>
```
marzer, line 10: See my note about `<algorithm>` earlier.

(Also I think the #if check for `TOML_HAS_CUSTOM_OPTIONAL_TYPE` is a mistake; copy paste?)

include/toml++/impl/node_view.h
```diff
@@ -719,6 +719,15 @@ TOML_NAMESPACE_START
...
 		}
 
+		/// \brief Returns a view of the subnode matching a fully-qualified "TOML path".
+		///
+		/// \see #toml::node::at_path(const toml::path&)
+		TOML_NODISCARD
+		node_view at_path(const toml::path& path) const noexcept
+		{
+			return node_ ? node_->at_path(path) : node_view{};
+		}
```
marzer, line 729: Node views probably want an `operator[]` overload for paths, too. Being able to mix regular strings and toml::paths when chaining `[]` on node views seems like a pretty potent mix!

include/toml++/impl/path.h
```diff
@@ -0,0 +1,411 @@
...
+
+		/// \brief	Copy constructor.
+		TOML_NODISCARD_CTOR
+		TOML_EXPORTED_MEMBER_FUNCTION
+		inline path(const path& other) { *this = other; }
+
+		/// \brief	Move constructor.
+		TOML_NODISCARD_CTOR
+		TOML_EXPORTED_MEMBER_FUNCTION
+		path(path&& other) noexcept { *this = std::move(other); }
```
marzer, line 96:
```suggestion
		/// \brief	Copy constructor.
		TOML_NODISCARD_CTOR
		inline path(const path& other) { *this = other; }

		/// \brief	Move constructor.
		TOML_NODISCARD_CTOR
		path(path&& other) noexcept { *this = std::move(other); }
```

TOML_EXPORTED_MEMBER_FUNCTION doesn't apply to inline functions ^_^

(also that move ctor can probably just be defaulted)

include/toml++/impl/path.h
```diff
@@ -0,0 +1,411 @@
...
+
+		/// \brief	Append operator.
+		TOML_EXPORTED_MEMBER_FUNCTION
+		path& operator+=(std::string_view);
+
+		/// \brief Evaluate whether path parsing succeeded
+		inline operator bool() const noexcept
+		{
+			return !parse_error_;
+		};
```
marzer, line 126:
```suggestion
		TOML_NODISCARD
		inline operator bool() const noexcept
		{
			return !parse_error_;
		};
```

Same for any "just gets a value and nothing else" functions - calling one of these types of functions and discarding the result is almost always an error, so `[[nodiscard]]` and friends help users catch bugs.

include/toml++/impl/path.h
```diff
@@ -0,0 +1,411 @@
...
+		return lhs + rhs;
+	}
+
+	inline path operator/(const wchar_t* lhs, const toml::path& rhs)
+	{
+		return lhs + rhs;
+	}
+
+#endif // TOML_ENABLE_WINDOWS_COMPAT
+
```
marzer, line 407: For non-member operators like this you almost always want to implement them in the class body and have them be `friend`; this makes them "[hidden friends](https://jacquesheunis.com/post/hidden-friend-compilation/)" which helps deal with ADL issues (+ potentially improve compile times in operator-overload-heavy scenarios)

marzer (changes requested):
Great work man! I am looking forward to releasing this feature. A few nits, but I've tried not to be too picky.

include/toml++/impl/path.inl
```diff
@@ -0,0 +1,521 @@
...
+#pragma once
+//# {{
+#include "preprocessor.h"
+#if !TOML_IMPLEMENTATION
+#error This is an implementation-only header.
+#endif
+//# }}
+
+#include "path.h"
+#include "std_algorithm.h"
```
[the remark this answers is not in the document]
jonestristand: It's me being extremely lazy and using std::clamp. It's a trivial function so I'll just implement it in the library

include/toml++/impl/path.h
```diff
@@ -0,0 +1,411 @@
...
+#include "std_string.h"
+#include "std_variant.h"
+
+/// \cond
+TOML_IMPL_NAMESPACE_START
+{
+
+}
+TOML_IMPL_NAMESPACE_END;
+/// \endcond
```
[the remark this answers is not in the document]
jonestristand: Absolutely correct, will remove!

include/toml++/impl/path.h
```diff
@@ -0,0 +1,411 @@
...
+TOML_NAMESPACE_START
+{
+
+	/// \brief Indicates type of path component, either a key, an index in an array, or invalid
+	enum class TOML_CLOSED_ENUM path_component_type : uint8_t
+	{
+		INVALID		= 0x0,
+		KEY			= 0x1,
+		ARRAY_INDEX = 0x2
+	};
```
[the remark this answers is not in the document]
jonestristand: 👍 👍

include/toml++/impl/path.h
```diff
@@ -0,0 +1,411 @@
...
+		bool operator!=(const char* compare) const noexcept;
+
+		/// \brief Fetch a path component by index
+		inline path_component* operator[](size_t index) noexcept
+		{
+			if (index < components_.size())
+				return &components_[index];
+
+			return nullptr;
+		};
```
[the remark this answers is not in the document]
jonestristand: That's a good point, and I can't remember why I decided to return a pointer here. It might have been to return nullptr if doesn't exist? Could just return an empty path_component with `.type = path_component_type::invalid` though.

Could certainly do some OOP stuff to maintain value/type relationship, though what I did with the variant is that at_path just checks that the contained type is sensible before returning, and if not returns an empty node_view.

include/toml++/impl/node_view.h
jonestristand: This is a very cool idea, I like it!

include/toml++/impl/path.h
jonestristand: Thanks, good catch!

include/toml++/impl/path.h
jonestristand: TIL!

include/toml++/impl/path.h
jonestristand: I had it set up this way initially but was having linker issues in test.cpp when trying to run the unit tests... unsure why, I'll move them back and have a closer look.

jonestristand:
Updates!

include/toml++/impl/path.h
```diff
@@ -0,0 +1,438 @@
...
+			return lhs + rhs;
+		}
+
+#endif // TOML_ENABLE_WINDOWS_COMPAT
+
+	};
+
+	inline namespace literals
+	{
+		TOML_ABI_NAMESPACE_BOOL(TOML_EXCEPTIONS, lit_ex, lit_noex);
```
marzer, line 406:
```suggestion
```

You don't need the inner ABI namespace here. That was specifically relating to whether or not the `toml::parse()` can throw, based on the setting for `TOML_EXCEPTIONS` - the ABI namespace in that scenario makes it so that people won't have weird errors if they accidently compile the library in one mode, and their application in another - instead it forces a hard linker error.

(again, something I've left undocumented and probably should note somewhere)

include/toml++/impl/path.h
```diff
@@ -0,0 +1,438 @@
...
+		/// \returns	\conditional_return{With exceptions}
+		///				A toml::table.
+		/// 			\conditional_return{Without exceptions}
+		///				A toml::parse_result.
+		TOML_NODISCARD
+		inline toml::path operator"" _tpath(const char* str, size_t len)
+		{
+			return toml::path(std::string_view{ str, len });
+		}
+		TOML_ABI_NAMESPACE_END; // TOML_EXCEPTIONS
```
marzer, line 433:
```suggestion
```

See note about ABI namespace above.

include/toml++/impl/path.h
```diff
@@ -0,0 +1,438 @@
...
+		}
+
+		friend std::istream& operator>>(std::istream& is, toml::path& rhs)
+		{
+			std::string s;
+			is >> s;
+			rhs.assign(s);
+
+			return is;
+		}
```
marzer, line 304: The implementations for these two functions should be moved to the `.inl` file since they'll require `#include <ostream>` and `#include <istream>`, something I only do in the `TOML_IMPLEMENTATION` parts of the library so that people who disable `TOML_HEADER_ONLY` can pay the cost of these includes only once. (I already `#include <iosfwd>` in `forward_declarations.h`)

Same goes for any functions that use stringstreams, if there are any.

Also, if you haven't seen it already, let me introduce you to a great "how expensive is this header" tool: https://artificial-mind.net/projects/compile-health/ - type "stream" into the search bar and be shocked :D

include/toml++/impl/path.inl
```diff
@@ -0,0 +1,498 @@
...
+#include "preprocessor.h"
+#if !TOML_IMPLEMENTATION
+#error This is an implementation-only header.
+#endif
+//# }}
+
+#include "path.h"
+
+TOML_DISABLE_WARNINGS;
+#include <sstream>
```
marzer, line 16:
```suggestion
#include <sstream>
#include <ostream>
#include <istream>
```

Per my note about these in the public header above.

include/toml++/impl/table.h
```diff
@@ -10,6 +10,7 @@
 #include "make_node.h"
 #include "node_view.h"
 #include "key.h"
+#include "path.h"
```
marzer, line 13: See my note about `#include "path.h"` above

marzer (changes requested):
Home stretch, methinks. Mostly just pedantic library/structural nits :)

include/toml++/impl/path.h
jonestristand: Terrifying to include filestream or complex or regex anywhere!

include/toml++/impl/at_path.inl
```diff
@@ -70,7 +70,7 @@ TOML_ANON_NAMESPACE_START
 #if TOML_INT_CHARCONV
 
 					auto fc_result = std::from_chars(index_str.data(), index_str.data() + index_str.length(), index);
-					if (fc_result.ec != std::errc{})
+					if (fc_result.ec != std::errc{} || fc_result.ptr != index_str.data() + index_str.length())
```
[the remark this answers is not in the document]
jonestristand: :-)

Comment added to no one has to go back to documentation in future ;-)

jonestristand:
Fixes pushed!

marzer:
Yup, looks good to me @jonestristand! There's a few bits-and-pieces I'll want to do post-merge so I won't merge it in until I have time to do that (this weekend mayyyybe, otherwise some weekend soon), so if there's more you think you want to add (any other utility functions etc.) feel free to push to this branch before then :)

marzer:
Going to look after this tomorrow or Sunday this weekend FYI @jonestristand

jonestristand:
Awesome, pumped to have it merged in! I'm just adding a subpath method today so luckily that'll sneak in under the wire ;-)

marzer:
Thanks @jonestristand! I'll keep it in the `paths` branch for a bit while I do some additional things, and will try to summarize changes I make in commit descriptions. Since you were interested in learning more and re-familiarizing yourself with C++, please let me know if you would like more detailed information about anything you see - I'll be happy to go into more detail on gitter :)

PR #161, opened 2022-06-20
feat(paths): Removed use of std::variant in path_component_value

jonestristand:
**What does this change do?**
Add placement new implementation for paths feature instead of std::variant

#153 

**Pre-merge checklist**

    - [ ] Clang 6 or higher
    - [ ] GCC 7 or higher
    - [x] MSVC 19.20 (Visual Studio 2019) or higher

### DIFF
--- a/include/toml++/impl/path.h
+++ b/include/toml++/impl/path.h
@@ -6,7 +6,6 @@
 
 #include "forward_declarations.h"
 #include "std_vector.h"
-#include "std_variant.h"
 #include "header_start.h"
 
 TOML_NAMESPACE_START
@@ -14,34 +13,62 @@ TOML_NAMESPACE_START
 	/// \brief Indicates type of path component, either a key, an index in an array, or invalid
 	enum class TOML_CLOSED_ENUM path_component_type : uint8_t
 	{
-		invalid		= 0x0,
 		key			= 0x1,
 		array_index = 0x2
 	};
-
-	/// \brief Holds the value of a path component, either the name of the key in a string_view, or the index of an array as a size_t
-	using path_component_value = std::variant<std::string, size_t>;
 
 	/// \brief Represents a single component of a complete 'TOML-path': either a key or an array index
 	struct TOML_EXPORTED_CLASS path_component
 	{
 		friend class path;
 
-	  private:
-		/// \cond
+		struct storage_t
+		{
+			static constexpr size_t size =
+				(sizeof(size_t) < sizeof(std::string) ? sizeof(std::string) : sizeof(size_t));
+			static constexpr size_t align =
+				(alignof(size_t) < alignof(std::string) ? alignof(std::string) : alignof(size_t));
 
-	  	path_component_value value_;
+			alignas(align) unsigned char bytes[size];
+		};
+		alignas(storage_t::align) mutable storage_t value_storage_;
+
 		path_component_type type_;
 
 		TOML_PURE_GETTER
 		TOML_EXPORTED_STATIC_FUNCTION
 		static bool TOML_CALLCONV equal(const path_component&, const path_component&) noexcept;
 
+		template <typename Type>
+		TOML_NODISCARD
+		TOML_ALWAYS_INLINE
+		static Type* get_as(storage_t& s) noexcept
+		{
+			return TOML_LAUNDER(reinterpret_cast<Type*>(s.bytes));
+		}
+
+		static void store_key(std::string_view key, storage_t& storage_)
+		{
+			::new (static_cast<void*>(storage_.bytes)) std::string{ key };
+		}
+
+		static void store_index(size_t index, storage_t& storage_)
+		{
+			::new (static_cast<void*>(storage_.bytes)) std::size_t{ index };
+		}
+
+		void destroy() noexcept
+		{
+			if (type_ == path_component_type::key)
+				get_as<std::string>(value_storage_)->~basic_string();
+		}
+
 		/// \endcond
 	  public:
-		/// \brief	Default constructor.
+		/// \brief	Default constructor (creates an empty key).
 		TOML_NODISCARD_CTOR
-		path_component() noexcept = default;
+		TOML_EXPORTED_MEMBER_FUNCTION
+		path_component();
 
 		/// \brief	Constructor for a path component that is an array index
 		TOML_NODISCARD_CTOR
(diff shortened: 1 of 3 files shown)


include/toml++/impl/path.inl
```diff
@@ -29,40 +29,122 @@ TOML_ENABLE_WARNINGS;
 TOML_NAMESPACE_START
 {
 	TOML_EXTERNAL_LINKAGE
-	path_component::path_component(size_t index) noexcept : value_(index), type_(path_component_type::array_index)
-	{ }
+	path_component::path_component() //
+		: type_{ path_component_type::key }
+	{
+		path_component::store_key("", value_storage_);
```
marzer, line 35: All of these `store_XXXX` calls shouldn't need the explicit `path_component::` prefix - _"the calls are coming from inside the house"_

include/toml++/impl/path.inl
```diff
@@ -29,40 +29,122 @@ TOML_ENABLE_WARNINGS;
...
+	TOML_EXTERNAL_LINKAGE
+	path_component& path_component::operator=(const path_component& rhs)
+	{
+		destroy();
+		type_ = rhs.type_;
+		if (type_ == path_component_type::array_index)
+			path_component::store_index(rhs.index(), value_storage_);
+		else
+			path_component::store_key(rhs.key(), value_storage_);
+		return *this;
```
marzer, line 90: Always pre-emptively destroying and re-instantiating is a pessimism; if LHS and RHS are both strings, the std::string's copy-assignment operator should be used instead (similar to what you're doing in the move-assign op)

include/toml++/impl/path.inl
```diff
@@ -29,40 +29,122 @@ TOML_ENABLE_WARNINGS;
...
 	{
-		return lhs.type_ == rhs.type_ && lhs.value_ == rhs.value_;
+		// Different comparison depending on contents
+		if (lhs.type_ != rhs.type_)
+			return false;
+
+		if (lhs.type_ == path_component_type::array_index)
+			return *get_as<size_t>(lhs.value_storage_) == *get_as<size_t>(rhs.value_storage_);
+		else // path_component_type::key
+			return *get_as<std::string>(lhs.value_storage_) == *get_as<std::string>(rhs.value_storage_);
```
marzer, line 126:
```suggestion
		if (lhs.type_ == path_component_type::array_index)
			return lhs.index() == rhs.index();
		else // path_component_type::key
			return lhs.key() == rhs.key();
```

include/toml++/impl/path.h
```diff
@@ -6,42 +6,69 @@
...
 		array_index = 0x2
 	};
 
-	/// \brief Holds the value of a path component, either the name of the key in a string_view, or the index of an array as a size_t
-	using path_component_value = std::variant<std::string, size_t>;
-
 	/// \brief Represents a single component of a complete 'TOML-path': either a key or an array index
 	struct TOML_EXPORTED_CLASS path_component
 	{
 		friend class path;
```
marzer, line 23:
```suggestion

```

Probably doesn't need to be a friend of `path` at all now that all the state handling has moved to the `path_component` itself

include/toml++/impl/path.inl
```diff
@@ -29,40 +29,122 @@ TOML_ENABLE_WARNINGS;
...
+	}
+
+	TOML_EXTERNAL_LINKAGE
+	path_component& path_component::operator=(path_component&& rhs) noexcept
+	{
+		destroy();
+
+		if (type_ != rhs.type_)
+		{
+			type_ = rhs.type_;
```
marzer, line 100:
```suggestion


		if (type_ != rhs.type_)
		{
			destroy();
			type_ = rhs.type_;
```

Otherwise the move-assign path would crash in the string case - you'd destroyed it first!

include/toml++/impl/path.inl
```diff
@@ -29,40 +29,122 @@ TOML_ENABLE_WARNINGS;
...
+		else // path_component_type::key
+			return *get_as<std::string>(lhs.value_storage_) == *get_as<std::string>(rhs.value_storage_);
 	}
 
 	TOML_EXTERNAL_LINKAGE
 	path_component& path_component::operator= (size_t index) noexcept
 	{
-		value_ = static_cast<size_t>(index);
-		type_  = path_component_type::array_index;
+		destroy();
```
marzer, line 132: As above (pre-emptive destroying). There's a few more instances of it, so I won't bother to spam the rest.

marzer (changes requested):
Looking pretty good! Just a few lifetime niggles to deal with and it will be good to go :)

include/toml++/impl/path.h
```diff
@@ -62,13 +89,130 @@ TOML_NAMESPACE_START
...
+		}
+
+		/// \brief	Returns the array index (rvalue overload).
+		TOML_NODISCARD
+		size_t&& index() && noexcept
 		{
-			return value_;
+			TOML_ASSERT_ASSUME(type_ == path_component_type::array_index);
+			return static_cast<size_t&&>(*get_as<size_t>(value_storage_));
 		}
```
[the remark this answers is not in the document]
jonestristand: Moved all non-const reference accessors to private methods, and changed the implicit casts to return const references

include/toml++/impl/path.h
```diff
@@ -6,42 +6,69 @@
...
-		invalid		= 0x0,
 		key			= 0x1,
 		array_index = 0x2
 	};
 
-	/// \brief Holds the value of a path component, either the name of the key in a string_view, or the index of an array as a size_t
-	using path_component_value = std::variant<std::string, size_t>;
-
 	/// \brief Represents a single component of a complete 'TOML-path': either a key or an array index
 	struct TOML_EXPORTED_CLASS path_component
```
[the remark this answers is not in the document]
marzer: _they grow up so fast_

include/toml++/impl/path.inl
jonestristand: Added to make explicit that these are static methods :-) Style stuff, will remove!

include/toml++/impl/path.inl
jonestristand: Interesting, when I step through this in debug, it doesn't actually crash. But your point is well taken and I've switched it!

include/toml++/impl/path.inl
jonestristand: Done

include/toml++/impl/path.inl
```diff
@@ -29,50 +29,154 @@ TOML_ENABLE_WARNINGS;
...
+			store_key(std::move(pc).key(), value_storage_);
+	}
+
+	TOML_EXTERNAL_LINKAGE
+	path_component& path_component::operator=(const path_component& rhs)
+	{
+		destroy();
+		if (type_ != rhs.type_)
+		{
+			destroy();
```
marzer, line 87:
```suggestion

		if (type_ != rhs.type_)
		{
			destroy();
```

Looks like that first `destroy()` was en error?

marzer (changes requested):

include/toml++/impl/path.inl
jonestristand: Oof yes. Weird, I remember deleting this destroy() call, must have Ctrl+Z one too many times or something.

PR #165, opened 2022-07-21
build(meson): Update path for .pc and .cmake files

diizzyy:
This changes install directories to more common ones:
.pc --> share/pkgconfig to lib/pkgconfig and libdata/pkgconfig on FreeBSD hosts
.cmake --> share/cmake to lib/cmake

### DIFF
--- a/include/meson.build
+++ b/include/meson.build
@@ -6,7 +6,6 @@ if not is_subproject
 	import('pkgconfig').generate(
 		name: meson.project_name(),
 		description: 'Header-only TOML config file parser and serializer for C++',
-		install_dir: get_option('datadir')/'pkgconfig',
 		url: 'https://marzer.github.io/tomlplusplus'
 	)
 endif
@@ -19,7 +18,7 @@ if get_option('generate_cmake_config') and not is_subproject
 	#cmake.write_basic_package_version_file(
 	#	name: meson.project_name(),
 	#	version: meson.project_version(),
-	#	install_dir: get_option('datadir')/'cmake'/meson.project_name(),
+	#	install_dir: get_option('libdir')/'cmake'/meson.project_name(),
 	#	arch_independent: true
 	#)
 	# In the meantime, install a pre-generated Package Version file
@@ -27,13 +26,13 @@ if get_option('generate_cmake_config') and not is_subproject
 		configuration: {'version': meson.project_version()},
 		input: '..'/'cmake'/'tomlplusplusConfigVersion.cmake.meson.in',
 		output: 'tomlplusplusConfigVersion.cmake',
-		install_dir: get_option('datadir')/'cmake'/meson.project_name()
+		install_dir: get_option('libdir')/'cmake'/meson.project_name()
 	)
 
 	cmake.configure_package_config_file(
 		name: meson.project_name(),
 		input: '..'/'cmake'/'tomlplusplusConfig.cmake.meson.in',
 		configuration: configuration_data({'includedir': get_option('includedir')}),
-		install_dir: get_option('datadir')/'cmake'/meson.project_name(),
+		install_dir: get_option('libdir')/'cmake'/meson.project_name(),
 	)
 endif


marzer:
Not really my area of expertise; @Tachi107, care to comment on these changes?

include/meson.build
```diff
@@ -3,12 +3,21 @@
 tomlplusplus_dep = declare_dependency(include_directories: include_dir)
 
 if not is_subproject
-	import('pkgconfig').generate(
+	if host_machine.system() == 'freebsd'
+		import('pkgconfig').generate(
 		name: meson.project_name(),
 		description: 'Header-only TOML config file parser and serializer for C++',
-		install_dir: get_option('datadir')/'pkgconfig',
+		install_dir: join_paths('libdata' , 'pkgconfig'),
```
eli-schwartz, line 10: You're repeating the same thing twice, which is pretty verbose, instead of defining `the_install_dir = join_paths(... , 'pkgconfig')` in your if, then passing that to pkgconfig.generate.

Anyway, if you remove the install_dir kwarg, Meson already handles freebsd for you and defaults to lib/pkgconfig on most systems, and libdata on FreeBSD.

include/meson.build
```diff
@@ -3,12 +3,21 @@
...
 		url: 'https://marzer.github.io/tomlplusplus'
-	)
+		)
+	else
+		import('pkgconfig').generate(
+		name: meson.project_name(),
+		description: 'Header-only TOML config file parser and serializer for C++',
+		install_dir: get_option('lib')/'pkgconfig',
```
eli-schwartz, line 17: "lib" isn't an option, but "libdir" is.

eli-schwartz (changes requested):
In general, this is a question of whether the information in these files contains architecture-dependent paths or architecture-independent paths. I believe this was originally implemented as is because includes on their own aren't architecture-dependent.

include/meson.build
diizzyy: Good to know, I managed to miss that section in the documentation. Thanks!

marzer:
Thanks @eli-schwartz. I'm happy with this change if you are.

@diizzyy ignore that `tipi.build` CI failure - that's nothing to do with your changes. It doesn't like PRs and I haven't bothered to figure out why.

marzer:
Thanks for your contribution @diizzyy!

Tachi107:
Meh, I honestly don't really like these changes... The files where installed into datadir because they are architecture independent (and toml++ is too), and moving them to libdir is less elegant IMO. But whatever, if BSDs do not support these common directories I guess this is fine. But please try fixing FreeBSD's pkg-config too

marzer:
Welp looks like I might have been premature in merging? Eh. I'll revisit this tomorrow.

diizzyy:
This isn't really (Free)BSD-dependant, datadir (share) isn't a "common" location for at least UNIX / UNIX-based systems if you're going to package software (see Debian,  Gentoo etc) so why default to it?

marzer:
Having slept on it, my thoughts on this are: I use unix-family OS-es about 20 minutes per week through WSL just for the sake of running unit tests so my knowledge in this area is basically zero and is unlikely to increase any time soon. I'll happily defer to whatever consensus arises if additional patches in this area are proposed ^_^

Tachi107:
> This isn't really (Free)BSD-dependant, datadir (share) isn't a "common" location for at least UNIX / UNIX-based systems if you're going to package software (see Debian,  Gentoo etc) so why default to it?

I'm not sure I get what you mean.

From the [FreeBSD handbook](https://docs.freebsd.org/en/books/handbook/basics/#dirstructure) I see that:

| Directory | Description |
| ---------- | -------- |
| **/usr/lib** | Archive libraries |
| **/usr/local** | Local executables and libraries. Also used as the default destination for the FreeBSD ports framework. Within /usr/local, the general layout sketched out by hier(7) for /usr should be used. Exceptions are the man directory, which is directly under /usr/local rather than under /usr/local/share, and the ports documentation is in share/doc/port. |
| **/usr/share** | Architecture-independent files. |

So, is seems that `share/` is indeed the directory where architecture independent files should be stored.

But I'm not a *BSD user, so I may be missing something.

Since you mentioned Debian, see for example [RapidJSON's file list](https://packages.debian.org/stable/all/rapidjson-dev/filelist) - the pkg-config file is installed under `/usr/share`

Tachi107:
> They specifically seem to want libdata and only libdata, but I don't know why. I thought I recalled from previous experience that they only replaced lib/pkgconfig with libdata/pkgconfig but apparently I am wrong...

Oh, interesting... Yeah, by looking at FreeBSD's pkgconf source it seems that they look for files in libdata. Now _that_ is uncommon, but I think we could handle that easily. I'm not sure Meson upstream should be fixed though

diizzyy:
That's true for the FreeBSD operating system, /usr/local is used for Ports/Packages (3rd party software that's not in "base" OS)
See https://www.freebsd.org/cgi/man.cgi?query=hier&sektion=7&format=html
There's a clear distinction between base and ports/packages

Looking at RapidJSON there are quite a few distros such as Alpine, Alpine, Fedora, Gentoo, OpenSUSE that doesn't use share but instead lib and/or lib64 so I guess Debian is the "oddball" in that regard

On top of my head if you look at https://repology.org/project/nlohmann-json/versions which I guess falls into the same category and is adopted by many distros it does follow the proposed change

Tachi107:
Il giorno ven 22 lug 2022 alle 04:04:29 -07:00:00, Daniel Engberg 
***@***.***> ha scritto:
> That's true for the FreeBSD operating system, /usr/local is used for 
> Ports/Packages (3rd party software that's not in "base" OS)
>  See 
> https://www.freebsd.org/cgi/man.cgi?query=hier&sektion=7&format=html
>  There's a clear distinction between base and ports/packages
> 
Yes, we're talking about /usr/local here.
> Looking at RapidJSON there are quite a few distros such as Alpine, 
> Alpine, Fedora, Gentoo, OpenSUSE that doesn't use share but instead 
> lib and/or lib64 so I guess Debian is the "oddball" in that regard
> 
Maybe because they don't care as much as Debian about wide CPU 
architecture support. Debian does, and that's why I try to make 
upstream use the appropriate dirs when possible.

But anyway, whether or not distros try to install arch-independent 
things to shared/ is irrelevant, because if they don't care why would 
they bother?

Upstream pkg-config explicitly says that "a typical Unix system will 
search in the directories /usr/lib/pkgconfig and /usr/share/pkgconfig" 
on their [guide][] and their [manpage][], and [OpenBSD's manpage][] 
says this too. The "oddball" that is ignoring share/ is FreeBSD's 
pkg-config; I'd really like to ask them why and what they recommend 
instead (like, should we use lib/ for arch-dependent libraries and 
libdata/ for arch-independent libs? Or is libdata/ always recommended?).

In the end, I don't see why Toml++ should always prefer installing to 
lib/ even when it is incorrect. Of course I could patch Debian's Toml++ 
package (that doesn't exist yet) to install to shared/, but I believe 
that the better solution would be to patch on FreeBSD instead, as they 
are the ones doing odd stuff.

[guide]: https://people.freedesktop.org/~dbn/pkg-config-guide.html#using
[manpage]: 
https://manpages.debian.org/testing/pkg-config/pkg-config.1.en.html
[OpenBSD's manpage]: https://man.openbsd.org/pkg-config.1

-- 
OpenPGP key: 66DE F152 8299 0C21 99EF  A801 A8A1 28A8 AB1C EE49

diizzyy:
I don't think it's enforced, having a quick look at Debian's packaging documentation there's no reference about it. Debian appears to use lib for nlohmann-json ( https://packages.debian.org/sid/all/nlohmann-json3-dev/filelist ) 
Another example of a header-only package that also install to lib
https://packages.debian.org/bookworm/amd64/librang-dev/filelist 

libdata is always used for pc files in FreeBSD, lib is however used on most Linux-based distros

If you look at packaged versions of tomlplusplus NetBSD installs to lib, yacp does the same

As a sidenote, the switch to pkgconf in many distros may also have played a part in this overall change

Tachi107:
> I don't think it's enforced, having a quick look at Debian's 
> packaging documentation there's no reference about it. Debian appears 
> to use lib for nlohmann-json ( 
> https://packages.debian.org/sid/all/nlohmann-json3-dev/filelist )
>  Another example of a header-only package that also install to lib
> https://packages.debian.org/bookworm/amd64/librang-dev/filelist
> 
I've never said that this is enforced, but it's cleaner and makes some 
things easier. I'm honestly not an expert when it comes to Debian's 
Multiarch rules and wide CPU coverage in general, but since I'm at 
DebConf I can ask an opinion of someone who surely knows better than us 
what's better to do and why.
> libdata is always used for pc files in FreeBSD, lib is however used 
> on most Linux-based distros
> 
Ok, thanks for the info :)
> If you look at packaged versions of tomlplusplus NetBSD installs to 
> lib, yacp does the same
> 
Yeah, this rule is not enforced. From the Debian policy:

 > The FHS requirement that architecture-independent 
application-specific static files be located in /usr/share is relaxed 
to a suggestion. In particular, a subdirectory of /usr/lib may be used 
by a package (or a collection of packages) to hold a mixture of 
architecture-independent and architecture-dependent files. However, 
when a directory is entirely composed of architecture-independent 
files, it should be located in /usr/share.

So in theory arch independent files _must_ be located in /usr/share on 
Linux, but Debian relaxes the rule as above. It's ok if you do it the 
"wrong" way, but doing things properly is certainly cleaner.
> As a sidenote, the switch to pkgconf in many distros may also have 
> played a part in this overall change
> 
Why? pkgconf also looks for stuff in share/

-- 
OpenPGP key: 66DE F152 8299 0C21 99EF  A801 A8A1 28A8 AB1C EE49

Tachi107:
Hey @diizzyy, I had a quick chat with a Debian Developer working on cross building, bootstrapping, MultiArch support, etc., and I've been told that installing .pc files to `/usr/lib` when they are related to architecture-independent libraries can indeed create issues when cross-building, as pkg-config cannot assume that that .pc file, residing in `/usr/lib`, is architecture independent, and so it won't be used for dependency resolution. The file can instead be always used for cross-builds when it resides in `/usr/share`, as in that case pkg-config can be sure that the library is arch-independent, and thus used for all architectures.

So, unless there's a good reason why toml++ would want to _always_ install .pc files in `/usr/lib`, I'd like this to be reverted. If FreeBSD really can't look for .pc files in `/usr/share` I guess that we could add an extra `if` to conditionally install to `libdata`, even though I'd prefer to fix the issue at the root, i.e. make FreeBSD look for stuff in `share/`.

diizzyy:
Since seemly everyone else doesn't enforce this it makes more sense to make an exception for Debian rather the other way around?

Tachi107:
> Since seemly everyone else doesn't enforce this

Untrue. The FHS does.

> and actually sets it to lib

I bet people do this because they don't think about this, not because they actually think that lib is for arch independent stuff.

> it makes more sense to make an exception for Debian rather the other way around?

No, it doesn't. I'm talking because about Debian here because that's what I know best, but everybody relying on the convention/rule that `share/` contains architecture-independent and `lib/` contains architecture-specific files could have issues with this patch.

If architecture-independent files have to go in `share/`, I don't understand why you really want to do otherwise. If you have a good technical reason, please say it ("FreeBSD does this so others should do it as well" is not a good technical reason, in my opinion).

diizzyy:
I've given you several examples where other distros (including Linux ones) does the do this yet you're quite adamant that it's a single OS. If you want to work against everyone sure go ahead, lets just say that adoption doesn't improve by doing so.

Tachi107:
> I've given you several examples where other distros (including Linux ones) and other projects doesn't do this

This is true, but if there is no reason why they do it I don't see why toml++ should follow them.

> yet you're quite adamant that it's a single OS. If you want to work against most distros packaging there's nothing stopping you but that does however seem counter productive?

I'm not sure I understand the first part of this, sorry. But yeah the only way to improve the way things work it to do them differently from how are currently done :) 

> Also, some packaging framework also allows you to set packages as architecture neutral

Debian allows this too, but pkg-config doesn't read packaging metadata, it only looks at the path and contents of .pc files

So, I'll repeat my question. Could you please give me a technical reason why architecture-independent should be installed to `lib/` instead of `share/`? If there is none, I don't see why are we arguing about this, we can simply install to `libdata/` when FreeBSD is detected

diizzyy:
I'll leave it up to you, I was just trying to make packaging easier for all parties.

Tachi107:
> I'll leave it up to you, I was just trying to make packaging easier for all parties.

Yeah, I understand. I'll think about this a bit more. I don't want to break stuff for non-Debian users, of course :D

eli-schwartz:
Clearly, /usr/share/pkgconfig is correct in at least some cases. An incredibly good example is bash-completion.pc, which specifies the `completionsdir=` directory for installing bash completion shell snippets to. In fact, that directory is usually in /usr/share as well, so it makes sense that the pkg-config file for a /usr/share directory is, also, in pkg-config's /usr/share hierarchy.

wayland-protocols also installs its pkg-config file there, which points on my system to `/usr/share/wayland-protocols` as the package datadir for getting system protocols (that `wayland-scanner` then converts from XML descriptions into generated C implementations while building some wayland-using project).

Maybe a bit more interesting is that e.g. xorg seems to use the share directory for `*proto.pc` (there's a bunch of them), apparently on the theory that anything which only provides Cflags pointing at the architecture-independent /usr/include contents, is not "lib" material, which aligns with what Debian seemingly recommends.

Other one-off pkg-config files in there include:
- yajl, which is a big oof, since it specifies `Libs: -lyajl`
- yaml-cpp, which is the same big oof

yaml-cpp also installs arch-dependent cmake files to the arch-independent directory /usr/share/cmake/yaml-cpp, go figure.

...

I would argue that pointing to what other projects do is not necessarily a proof for how *you* should do anything. You'll find plenty of examples out there that run the gamut from good to bad with stopovers at "indifferent" and "not applicable".

Tachi107:
Asked about this on the freebsd-hackers mailing list, linking here as a cross-reference: https://lists.freebsd.org/archives/freebsd-hackers/2022-July/001347.html

Tachi107:
The FreeBSD pkgconf maintainer explained that:

- FreeBSD only looks for .pc files in `libdata/` (@eli-schwartz, does Meson already account for this?)
- FreeBSD packaging (ports) automatically moves .pc files from `lib/` and `share/` yo `libdata/`, so reverting this wouldn't cause issues for them

Thanks to everyone involved, I've learnt something new about FreeBSD :)

@marzer, could you please revert this commit? (unless I have missed something and @diizzyy has some more concerns)

Some more useful references:

- How FreeBSD currently does things: https://lists.freebsd.org/archives/freebsd-hackers/2022-July/001349.html
- Why FreeBSD doesn't look in `share/`: https://lists.freebsd.org/archives/freebsd-hackers/2022-July/001350.html

eli-schwartz:
> FreeBSD only looks for .pc files in `libdata/` (@eli-schwartz, does Meson already account for this?)

Only for the case of `pkgconfig.generate(..., dataonly: false)`, which is the default.

diizzyy:
Ports framework doesn't move from share

Tachi107:
Uhm, this is probably something that has been overlooked by some FreeBSD dev, but yeah until we find a solution (that might involve fixing ports / changing pkgconf's behaviour) it is better to keep things as they are

diizzyy:
meson does it right by default so how is that "wrong"?

Tachi107:
> meson gets it right by default (like it is right now) so how is that "wrong"?

I mean that ports should also move .pc file from `share/` to `libdata/`, unless FreeBSD's pkgconf starts recognizing `share/pkgconfig/` as a valid pkg-config directory

> Edit: nm, CMake only install .cmake files no .pc at all

Not sure what you mean by this

marzer:
Sorry folks, been away from this for a bit. Have we reached a consensus here? Do I need to revert?

Tachi107:
In theory FreeBSD automatically adjusts paths of .pc and .cmake files automatically, in practice it doesn't do it for share/.

This should be solved in FreeBSD (either by making ports move files from share/ or by making FreeBSD's pkgconf look into share/), so toml++ shouldn't care IMO, but until FreeBSD doesn't deal with this inconsistency you can leave things as they are now.

@diizzyy, could you please try to ping the FreeBSD guys again?

diizzyy:
This (which I've explained several times by now) applies to more distros/OSes than just FreeBSD, is there any practical gain except that it generates more work when packaging?

marzer:
> which I've explained several times by now

Simmer down there chief. I don't have a horse in this race, and it appears @Tachi107 has swung around to keeping it as you've changed it. I just want a confirmation.

Tachi107:
> This (which I've explained several times by now) applies to more distros/OSes than just FreeBSD

Could you please provide an example? I did not find any (I've not searched a lot, but still).

> is there any practical gain except that it generates more work when packaging?

Yep: correctness and [cross-building](https://github.com/marzer/tomlplusplus/pull/165#issuecomment-1193146210). And in your case it generates more work when packaging because Ports only moves .pc files from `lib/` to `libdata/`, but somebody forgot about `share/`.

And by the way, we've been talking about .pc for the most part, while you also modified the location of .cmake files; [CMake's documentation](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) explicitly says that `share/` is always checked, so there's no reason why you would want to move them to `lib/`.

eli-schwartz:
> This (which I've explained several times by now) applies to more distros/OSes than just FreeBSD

There's two bullet points to discuss here:
- the OS known as FreeBSD uses libdata not share or lib
- packages may be arch-dependent (libdata on FreeBSD, lib on Linux distros) or arch-independent (still libdata on FreeBSD, share on Linux distros)

You've explained several times that the choice of where to install to applies to more *projects* than tomlplusplus; you have not explained that libdata applies to more *distros* than just FreeBSD.

On an unrelated note, you've also explained that only Debian maintainers go around asking projects with arch-independent content to be careful to put their pkg-config files in the arch-independent directory, although I'm not sure whether you did a proper poll of this? I don't see why it's relevant, anyway.

It's semantically correct to use arch-independent, but it may not be "worth the fuss", the only penalty for doing it semantically wrong is that people using tomlplusplus have to have one copy for each cross-compilation target, assuming they support cross-compilation natively to begin with.

diizzyy:
I didn't, Tachi107 did (regarding Debian), I just gathered a small amount of real life data.

NetBSD uses lib not libexec for example but at this point it seems rather meaningless to continue this discussion, just let distros and/package managers repackage it as they see fit and do whatever changes you seem git even if it seemingly adds more work for most people.

Tachi107:
FreeBSD's pkgconf now looks for files in share/, this change can now be reverted 🥳

Thanks to everyone involved :)

https://cgit.freebsd.org/ports/commit/?id=d48fab59daa56e0b3b6ffecef57a69c32ae9c0a7

PR #166, opened 2022-07-29
Revert "build(meson): Update path for .pc and .cmake files (#165)"

Tachi107:
If you want to read the full story please see pull request #165, but in short moving files from `share/` to `lib/` was needed because FreeBSD didn't look for .pc files in `share/pkgconfig/`.

After a quick [discussion][] upstream, FreeBSD's [pkgconf was patched][] to also honour .pc files in `share/pkgconfig/`.

Now that everybody looks for stuff in `share/`, the commit b1ecdf0ed86faeeaebe13d0db16ad04c20519527 can be reverted, so that stuff like cross-compilation can continue working fine.

[discussion]: https://lists.freebsd.org/archives/freebsd-hackers/2022-July/001355.html
[pkgconf was patched]: https://github.com/freebsd/freebsd-ports/commit/d48fab59daa56e0b3b6ffecef57a69c32ae9c0a7

### DIFF
--- a/include/meson.build
+++ b/include/meson.build
@@ -6,6 +6,7 @@ if not is_subproject
 	import('pkgconfig').generate(
 		name: meson.project_name(),
 		description: 'Header-only TOML config file parser and serializer for C++',
+		install_dir: get_option('datadir')/'pkgconfig',
 		url: 'https://marzer.github.io/tomlplusplus'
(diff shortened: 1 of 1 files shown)


marzer:
Thanks @Tachi107 :)

PR #171, opened 2022-09-11
Suppress -Wduplicated-branches

HazardyKnusperkeks:
Solves:
tomlplusplus/include/toml++/impl/path.h:29:73: warning: this condition has identical branches [-Wduplicated-branches]
   29 |                                 (alignof(size_t) < alignof(std::string) ? alignof(std::string) : alignof(size_t));

**What does this change do?**
Fix a Warning.

Nope

**Pre-merge checklist**

- [X] I've read [CONTRIBUTING.md]
- [X] I've rebased my changes against the current HEAD of `origin/master` (if necessary)
- [ ] I've regenerated toml.hpp ([how-to]) - Could not, the pip command does not run successfully, and no energy to fix that.
- [X] I've updated any affected documentation
    - [ ] Clang 6 or higher
    - [ ] GCC 7 or higher

### DIFF
--- a/include/toml++/impl/preprocessor.h
+++ b/include/toml++/impl/preprocessor.h
@@ -577,6 +577,7 @@
 	TOML_PRAGMA_GCC(diagnostic ignored "-Wmaybe-uninitialized")                                                        \
 	TOML_PRAGMA_GCC(diagnostic ignored "-Wnoexcept")                                                                   \
 	TOML_PRAGMA_GCC(diagnostic ignored "-Wnull-dereference")                                                           \
+	TOML_PRAGMA_GCC(diagnostic ignored "-Wduplicated-branches")                                                        \
 	static_assert(true)
 
 #define TOML_POP_WARNINGS                                                                                              \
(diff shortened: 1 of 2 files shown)


marzer (changes requested):
Thanks for the contribution, however `std::max` requires `<algorithm>` which I have deliberately avoided including (because it is absolutely enormous). There's an internal `min()` function in the `impl` namespace for this purpose (`forward_declarations.h`) - re-write this in terms of that function and it will be good to go!

Alternatively, I'd also be happy with simply suppressing the warning because it's clearly spam in this case. Adding it to the appropriate `TOML_DISABLE_SPAM_WARNINGS` macro in `preprocessor.h` would be the right move there.

HazardyKnusperkeks:
> Thanks for the contribution, however `std::max` requires `<algorithm>` which I have deliberately avoided including (because it is absolutely enormous). There's an internal `min()` function in the `impl` namespace (`forward_declarations.h`) - if you add an equivalent `max()` there and use that instead it will be good to go!
> 
> Alternatively, I'd also be happy with simply suppressing the warning because it's clearly spam in this case. Adding it to the appropriate `TOML_DISABLE_SPAM_WARNINGS` macro in `preprocessor.h` would be the right move there.

Done.

marzer:
Thanks!

PR #173, opened 2022-09-18
Fix typo in at_path

a-is:
**What does this change do?**
Using a static code analyzer, a suspicious place was discovered:
```cpp
if (c == ']')
{
    pos++;
    break;
}
else if TOML_UNLIKELY(c == '.' || c == '[')
//                    ^^^^^^^^
    break;
else if (c == '\t' || c == '.')
//                    ^^^^^^^^
    pos++;
else
    return false;
```

Obviously, a mistake was made in one of the places. It is easy to see that in the second case, whitespace characters are checked, so there should be a space (`' '`) instead of a dot (`'.'`). This PR fixing bug, adds a test.

No, I created a PR without creating an issue

**Pre-merge checklist**

    - [ ] Clang 6 or higher
    - [x] GCC 7 or higher


marzer:
Nice find, thanks!

PR #238, opened 2024-10-17
meson: install header and pkg-config file and in subprojects

stephanlachnit:
**What does this change do?**

Install the headers and the pkg-config file also in Meson subprojects. The reasoning behind this is that if you install a Meson project using TOML++ (potentially into a non-standard location like `/opt/project`), and also create a pkg-config file for your project, TOML++ might get added as public or private dependency. In that case, the header (and pkg-config file) should be installed.

No

**Pre-merge checklist**

-   [ ] ~I've added new test cases to verify my change
    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/include/meson.build
+++ b/include/meson.build
@@ -5,22 +5,18 @@
 
 include_dir = include_directories('.')
 
-if not is_subproject
-	install_subdir('toml++', install_dir: get_option('includedir'))
-endif
+install_subdir('toml++', install_dir: get_option('includedir'))
 
 if not build_lib # header-only mode
 
 	tomlplusplus_dep = declare_dependency(include_directories: include_dir)
 
-	if not is_subproject
-		import('pkgconfig').generate(
-			name: meson.project_name(),
-			description: 'Header-only TOML config file parser and serializer for C++',
-			install_dir: get_option('datadir')/'pkgconfig',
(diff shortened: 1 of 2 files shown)


marzer:
> The reasoning behind this is that if you install a Meson project using TOML++  [...] TOML++ might get added as public or private dependency.

The word "might" is the operative one here - in my experience most subproject usage is just to bring in a dependency for internal use, so this is likely going to be problematic for many (me included - this would complicate one of my workflows on Windows).

I'd happy accept a version of these changes that left the current behaviour as default, and added some meson option to enable all the install-ish bits even as a subproject, though. I think that's a good middle-ground.

stephanlachnit:
> this is likely going to be problematic for many (me included - this would complicate one of my workflows on Windows).

I'm not entirely sure how this would complicate any workflow - it just is an additional header and file? I guess on Windows you most likely don't want to install any header anyway, in which case [installation tags](https://mesonbuild.com/Installing.html#installation-tags) are the better option any.

marzer:
> I'm not entirely sure how this would complicate any workflow

Welcome to hyrum's law.

As noted above, I won't accept this if you change default behaviour. Please modify it per my request above, or close your PR. Thanks.

PR #244, opened 2025-01-30
let `print_to_stream` also print the end of a source region

N-Dekker:
With this commit, `std::cout << node.source()` should print both the begin and the end of the source region.

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/include/toml++/impl/print_to_stream.inl
+++ b/include/toml++/impl/print_to_stream.inl
@@ -419,8 +419,11 @@ TOML_IMPL_NAMESPACE_START
 	void TOML_CALLCONV print_to_stream(std::ostream & stream, const source_region& val)
 	{
 		print_to_stream(stream, val.begin);
-		print_to_stream(stream, " to "sv);
-		print_to_stream(stream, val.end);
+		if (val.begin != val.end)
+		{
+			print_to_stream(stream, " to "sv);
+			print_to_stream(stream, val.end);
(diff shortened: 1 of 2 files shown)


include/toml++/impl/print_to_stream.inl
```diff
@@ -419,6 +419,8 @@ TOML_IMPL_NAMESPACE_START
 	void TOML_CALLCONV print_to_stream(std::ostream & stream, const source_region& val)
 	{
 		print_to_stream(stream, val.begin);
+		print_to_stream(stream, " to "sv);
+		print_to_stream(stream, val.end);
```
marzer, line 423: There are situations wherein a source region can span zero characters, so it might be useful to take this into account:

```suggestion
		if (val.begin != val.end)
		{
			print_to_stream(stream, " to "sv);
			print_to_stream(stream, val.end);
		}
```

N-Dekker:
@marzer Thank you for the suggestion! Shall I squash you suggested commit on my fork? Otherwise you may also do a squash-and-merge on GitHub when you think this PR is good enough, of course.

marzer:
All good, we can squash-and-merge when it is brought into main. Just need to update the single-header version and I'll merge it in 👍🏼

marzer:
Thanks 🎉

Issue #219, opened 2024-01-19
Iterate over nested Tables (Array of Tables)


KevDi:
Hello, this is more a question then a real bug but there was no question to select.

I'm currently try to figure out how to iterate over a set of nested Tables. I tried the example from the Docs for the Arrays but this seems not to work, because i got compiler errors.

This is the Code that i tried:

```
int main() {
    toml::table tbl;
    try {
        tbl = toml::parse_file("config.toml");
        std::cout << tbl["data"] << '\n';
        toml::array* pipes = tbl["data"].as_array();
        pipes->for_each([](auto&& ele) {
            std::cout << ele << '\n';
        });

    } catch (const toml::parse_error& error) {
        std::cout << error << '\n';
    }

    return 0;
}
``` 

And these are the errors i got:

```
error C2338: static_assert failed: 'TOML node visitors must be invocable for at least one of the toml::node specializations: ...
error C3313: 'keep_going': variable cannot have the type 'const void' ...
error C3536: 'keep_going': cannot be used before it is initialized ...
```

The Toml i used:

```
[[data]]
name = "Data_One"

[[data]]
name = "Data_two"
```

With the following Code i was able to iterate over the nested Tables but i wonder if this is the best or recommended way to do it.

```
int main() {
    toml::table tbl;
    std::vector<Pipeline> pipes;
    try {
        tbl = toml::parse_file("config.toml");
        std::cout << tbl["data"] << '\n';
        auto pipe = tbl["data"].as_array();
        for (auto& ele : *pipe) {
            std::cout << "----\n";
            std::cout << *ele.as_table() << '\n';
            std::cout << "----\n";
        }

    } catch (const toml::parse_error& error) {
        std::cout << error << '\n';
    }

    return 0;
}
```

marzer:
Yeah for_each() should work in the way you've tried it in the first example, and it's not immediately clear to me why it's not here. I'll have a look into it a bit later :)

KevDi:
If it helps this is the complete Error Message i got using the `for_each`:

```
C:\Users\kev\.conan2\p\tomlp86c984de8935d\p\include\toml++\impl\array.hpp(954,17): error C2057: expected constant expression [C:\Users\kev\Documents\Unterlagen\Projekte\programming\cpp\_vs\LogAndConfig\
build\LogAndConfig.vcxproj]
  (compiling source file '../src/main.cpp')
[...]

PR #247, opened 2025-02-06
add TOML_DISABLE_CONDITIONAL_NOEXCEPT_LAMBDA workaround MSVC error C2057

N-Dekker:
Offered a workaround to compile errors like "error C2057: expected constant expression", when using the "legacy lambda processor" of Visual C++. Such compile errors were reported by Kevin Dick, Jan 19, 2024, at issue #219

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/include/toml++/impl/array.hpp
+++ b/include/toml++/impl/array.hpp
@@ -951,7 +951,10 @@ TOML_NAMESPACE_START
 					static_cast<node_ref>(static_cast<Array&&>(arr)[i])
 						.visit(
 							[&]([[maybe_unused]] auto&& elem) //
+// Define this macro as a workaround to compile errors caused by a bug in MSVC's "legacy lambda processor".
+#if !TOML_DISABLE_CONDITIONAL_NOEXCEPT_LAMBDA
 							noexcept(for_each_is_nothrow_one<Func&&, Array&&, decltype(elem)>::value)
+#endif
 							{
 								using elem_ref = for_each_elem_ref<decltype(elem), Array&&>;
 								static_assert(std::is_reference_v<elem_ref>);
--- a/README.md
+++ b/README.md
@@ -290,6 +290,7 @@ UTF-8 decoding is performed using a state machine based on Bjoern Hoehrmann's '[
 -   **[@levicki](https://github.com/levicki)** - Helped design some new features
 -   **[@moorereason](https://github.com/moorereason)** - Reported a whole bunch of bugs
 -   **[@mosra](https://github.com/mosra)** - Created the awesome [m.css] used to generate the API docs
+-   **[@N-Dekker](https://github.com/N-Dekker)** - Added a workaround for the legacy lambda processor of MSVC 2019/2022
 -   **[@ned14](https://github.com/ned14)** - Reported a bunch of bugs and helped design some new features
 -   **[@okureta](https://github.com/okureta)** - Reported a bug
 -   **[@prince-chrismc](https://github.com/prince-chrismc)** - Added toml++ to ConanCenter, and fixed some typos
--- a/include/toml++/impl/preprocessor.hpp
+++ b/include/toml++/impl/preprocessor.hpp
@@ -1241,6 +1241,17 @@ TOML_ENABLE_WARNINGS;
 /// \detail Defaults to `0`.
 //# }}
 
+#ifndef TOML_DISABLE_CONDITIONAL_NOEXCEPT_LAMBDA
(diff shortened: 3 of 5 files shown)


include/toml++/impl/array.hpp
```diff
@@ -951,7 +951,12 @@ TOML_NAMESPACE_START
 					static_cast<node_ref>(static_cast<Array&&>(arr)[i])
 						.visit(
 							[&]([[maybe_unused]] auto&& elem) //
+// Define this macro as a workaround to MSVC 2019/2022 "error C2057: expected constant expression",
+// as reported by Kevin Dick, Jan 19, 2024, at https://github.com/marzer/tomlplusplus/issues/219
+// These compile errors were caused by a bug in the old "legacy lambda processor" of Visual C++.
+#ifndef TOML_DISABLE_CONDITIONAL_NOEXCEPT_LAMBDA
```
marzer, line 957: So, if you have a look in the ["user defines" section of preprocessor.hpp](https://github.com/marzer/tomlplusplus/blob/d00dd1f5cbc53de242dc75d1d1e9fdca11ecb041/include/toml%2B%2B/impl/preprocessor.hpp#L906-L1244) you'll see that all the "bool-like" `#define` compiler switches are implemented in terms of them being zero or non-zero, not simply defined or undefined.

This is based on a personal preference of mine; I don't like #define-based feature tests because they allow this logical inconsistency:

```
#define FOO_ENABLE_MAGIC_FEATURE 0
```

The user has defined it, but it's zero; if the feature test is implemented in terms of simply #define, this turns MAGIC_FEATURE _on_ 😅 

For for this new switch to fit in with the rest of them, you'd need to:

1. have this in the user defines section of preprocessor.hpp:

```
#ifndef TOML_DISABLE_CONDITIONAL_NOEXCEPT_LAMBDA
#define TOML_DISABLE_CONDITIONAL_NOEXCEPT_LAMBDA 0
#endif
```

2. change the logical tests to be simply `#if !TOML_DISABLE_CONDITIONAL_NOEXCEPT_LAMBDA`

marzer:
Oh, hmmn, also it seems like my CI setup in this repo has gotten stale 😅 Guess I should address whatever it's complaining about.

marzer:
Oh, actually, here's something else useful: MSVC actually gives you a way to detect that you're using the old or new preprocessor, per https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170:

image

It may be sensible to apply this detection to use it to set the default value of your new `TOML_DISABLE_CONDITIONAL_NOEXCEPT_LAMBDA`

N-Dekker:
> Oh, actually, here's something else useful: MSVC actually gives you a way to detect that you're using the old or new preprocessor

Interesting, but does that also detect which lambda processor is being used? (The compile errors are caused by the lambda processor, not the preprocessor.)

marzer:
Oh, yeah, never mind. I was mixing up _pre_-processor with _lambda_ processor. Do'h. Doesn't look like they have any magic macro for indicating the lambda processor that's in use.

include/toml++/impl/array.hpp
```diff
@@ -951,7 +951,10 @@ TOML_NAMESPACE_START
 					static_cast<node_ref>(static_cast<Array&&>(arr)[i])
 						.visit(
 							[&]([[maybe_unused]] auto&& elem) //
+// Define this macro as a workaround to compile errors caused by a bug in MSVC's "legacy lambda processor".
+#if !TOML_DISABLE_CONDITIONAL_NOEXCEPT_LAMBDA
 							noexcept(for_each_is_nothrow_one<Func&&, Array&&, decltype(elem)>::value)
```
N-Dekker, line 956: For what it's worth, I did try to work around the problem (locally) by introducing a _variable template_ `for_each_is_nothrow_one_v`, in order to do:

```cpp
    noexcept( for_each_is_nothrow_one_v<Func&&, Array&&, decltype(elem)> )
```
But it still did not make the old legacy lambda processor happy. 🤷

N-Dekker:
FYI, The legacy lambda processor errors are still there "by default", when generating a vcxproj file for a C++17 build by the latest version of CMake (version 3.31.5) _and_ using the latest VS2022 update (version 17.12.4). Specifically, CMake generates project files that have _Conformance mode_ = _Default_, which implies using legacy lambda processor.

marzer:
@Smjert was kind enough to fix the various failing CI issues for me @N-Dekker so if you want to rebase we should see some proper CI results 😄

marzer (approved):
Thanks @N-Dekker!

PR #250, opened 2025-02-12
ci: Fix python collision with system installation

Smjert:
**What does this change do?**

~- Install a python version using the actions/setup-python action, to have more control and consistency on the version used across workflows.~

- Install pip packages as a user instead of root, to avoid colliding with or messing up the system.

**Pre-merge checklist**

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,14 +57,14 @@ jobs:
             - name: Install base dependencies
               run: |
                   sudo apt -y update
-                  sudo apt -y install --no-install-recommends git python3 python3-pip ninja-build libstdc++-${{ env.gcc_version }}-dev locales-all
+                  sudo apt -y install --no-install-recommends git ninja-build libstdc++-${{ env.gcc_version }}-dev locales-all
 
             - uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


marzer:
RE the setup-python job, it's probably not even necessary. I don't know why I was bothering. For meson the system-default python should be fine.

Smjert:
Hello @marzer, using the setup-python actions is mostly a suggestion to take control of which version (at least the major) is used, and to have that be consistent across systems.

That being said, it's true that you only use python for `meson` itself so it's not strictly necessary. In the case of Ubuntu, the runners come with an already installed version from the repository (a 3.12), plus they also cache additional versions on a separate folder, a mix of newer and older ones. These are the ones that are selected by `setup-python`, if you don't specify a precise version or if you don't use the `check-latest: true` option.

In any case if you don't care too much of which version is used, you can indeed just use what's available, and the important part is `--user`, to not touch the system installed python packages.

marzer:
> Let me know what you want to do.

Yeah, I think just using what's on the runner is the preferred move.

marzer:
Well, looks good to me, thanks! 🎉

Issue #254, opened 2025-02-20
Utility functions to get a line or region of an input document as a string


N-Dekker:
**Is your feature request related to a problem? Please describe.**

When the parser encounters an error, or when the TOML input document is not according to application specific requirements, it would be helpful to the end user to present the line or the source region that appears problematic.  The `source()` member functions of `toml::parse_error` and `toml::node` tell us where to find the specific source region, but they don't yet give us the region as a string.

**Describe the solution you'd like**

It would be very helpful to have utility functions like `get_line(doc, line_num)` and `get_region(doc, begin_position, end_position)`, returning the specific line or region from the specified document as a string (or as a `string_view`).  

`get_line` might be implemented as follows:

```cpp
std::string_view get_line(std::string_view doc, source_index line_num)
{
	if (line_num == 0)
	{
		// Invalid line number. Should be greater than zero.
		return {};
	}

	// The position of the first character of the specified line.
	const auto begin_of_line = [doc, line_num]() -> std::size_t
	{
		if (line_num == 1)
		{
			return 0;
		}

		const auto num_chars = doc.size();
		std::size_t current_line_num{ 1 };

		for (std::size_t i{}; i < num_chars; ++i)
		{
			if (doc[i] == '\n')
			{
				++current_line_num;

				if (current_line_num == line_num)
				{
					return i + 1;
				}
			}
		}
		return std::string_view::npos;
	}();

	if (begin_of_line >= doc.size())
	{
		return {};
	}

	const auto end_of_line = doc.find_first_of("\r\n", begin_of_line);

	return doc.substr(begin_of_line,
					  (end_of_line == std::string_view::npos) ? std::string_view::npos : (end_of_line - begin_of_line));
}

```

**Additional context**

For [elastix ](https://elastix.dev) I already wrote a similar function, named `GetLine`: https://github.com/SuperElastix/elastix/blob/a73dabcc60eaee783089e4e611562bbd77fb1ac7/Common/ParameterFileParser/itkParameterFileParser.cxx#L307-L350  However, I believe that such a function could be useful for others as well, and I would like to prevent reinventing the wheel.

marzer:
[...]

PR #255, opened 2025-02-21
add `get_line(doc, line_num)` utility

N-Dekker:
**What does this change do?**

This pull request adds a utility function, `get_line(doc, line_num)`, which returns the specified line as a `string_view`.

This feature was discussed at issue #254.

**Pre-merge checklist**

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/include/toml++/impl/source_region.hpp
+++ b/include/toml++/impl/source_region.hpp
@@ -217,6 +217,78 @@ TOML_NAMESPACE_START
 			return lhs;
 		}
 	};
+
+	/// \brief	Returns the line at the specified line number, from the specified document.
+	///
+	/// \detail \cpp
+	/// auto doc = "alpha = 1\nbeta = 2"sv;
+	/// auto second_line = toml::get_line(doc, 2);
+	///
+	/// std::cout << "The second line says \"" << second_line << "\"\n";
+	/// \ecpp
+	///
+	/// \out
+	/// The second line says "beta = 2"
+	/// \eout
+	///
+	/// \param 	doc	The document.
+	/// \param 	line_num	The line number (1-based).
+	///
+	/// \returns	The specified line, excluding any possible trailing carriage return or line feed character.
+	/// \remarks Returns an empty string_view when there is no line at the specified line number, in the specified document.
+	TOML_NODISCARD
+	inline std::string_view get_line(std::string_view doc, source_index line_num) noexcept
+	{
+		if (line_num == 0)
+		{
+			// Invalid line number. Should be greater than zero.
+			return {};
+		}
+
+		// The position of the first character of the specified line.
+		const auto begin_of_line = [doc, line_num]() -> std::size_t
+		{
+			if (line_num == 1)
+			{
+				return 0;
+			}
+
+			const auto num_chars_of_doc = doc.size();
+			std::size_t current_line_num{ 1 };
+
+			for (std::size_t i{}; i < num_chars_of_doc; ++i)
+			{
+				if (doc[i] == '\n')
+				{
+					++current_line_num;
+
+					if (current_line_num == line_num)
+					{
+						return i + 1;
+					}
+				}
+			}
+			return std::string_view::npos;
+		}();
+
+		if (begin_of_line >= doc.size())
+		{
+			return {};
+		}
+
+		if (const auto end_of_line = doc.find('\n', begin_of_line); end_of_line != std::string_view::npos)
+		{
+			const auto num_chars_of_line = end_of_line - begin_of_line;
+
+			// Trim an optional trailing carriage return.
+			return doc.substr(begin_of_line,
+							  ((num_chars_of_line > 0) && (doc[end_of_line - 1] == '\r')) ? num_chars_of_line - 1
+																						  : num_chars_of_line);
+		}
+
+		// Return the last line. Apparently this doc has no trailing line break character at the end.
+		return doc.substr(begin_of_line);
+	}
 }
 TOML_NAMESPACE_END;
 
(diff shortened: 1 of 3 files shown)


include/toml++/impl/source_region.hpp
```diff
@@ -217,6 +217,70 @@ TOML_NAMESPACE_START
...
+	/// \out
+	/// The second line says "beta = 2"
+	/// \eout
+	///
+	/// \param 	doc	The document.
+	/// \param 	line_num	The line number (1-based).
+	///
+	/// \returns	The specified line, excluding any possible trailing carriage return or line feed character.
+	/// \remarks Returns an empty string_view when there is no line at the specified line number, in the specified document.
+	inline std::string_view get_line(std::string_view doc, source_index line_num)
```
marzer, line 239: Looks good to me, only thing I'd want to address before accepting this contribution would be the `inline`-ness of it. The library supports traditional compilation as static/shared - this function being a regular old non-template free function makes it a perfect canditate to be moved to the "source" section.

Because `toml++` is also header-only by default, the way this works is a bit non-obvious. The changes you'd need to make would consist of three main steps:

1. Change this to a declaration with the right decorators, like:
```suggestion
	TOML_NODISCARD
	TOML_EXPORTED_FREE_FUNCTION
	std::string_view get_line(std::string_view doc, source_index line_num);
```

2. Create an ".inl" file to host the implementation (see `impl/at_path.inl` for implementation guidance)
3. Add an include for the implementation part in toml.hpp ([examples](https://github.com/marzer/tomlplusplus/blob/cbc00d6766f14bcd2b3e72de2a3c369a0d48b98a/include/toml%2B%2B/toml.hpp#L60-L76))

Of course I can do these parts for you too, if you like.

include/toml++/impl/source_region.hpp
```diff
@@ -217,6 +217,70 @@ TOML_NAMESPACE_START
...
+			}
+			return std::string_view::npos;
+		}();
+
+		if (begin_of_line >= doc.size())
+		{
+			return {};
+		}
+
+		const auto end_of_line = doc.find_first_of("\r\n", begin_of_line);
```
marzer, line 278: Also I think this should just be `\n`:

```suggestion
		const auto end_of_line = doc.find_first_of("\n", begin_of_line);
```

marzer:
On an unrelated note, I just checked out elasix, and that's a cool project! My previous job related to physics simulation for medical training purposes and we were starting to investigate populating simulations based on imaging data when I left. Looks like it may have come in handy.

include/toml++/impl/source_region.hpp
marzer: The TOML spec would regard that as being just one line, so the same thing it would return for `"embedded X here"`.

> as it seemed like an exotic and irrelevant scenario to me

For your use-case, yes. But if it's in the library proper, I don't get the luxury of thinking like that unfortunately 😅

include/toml++/impl/source_region.hpp
N-Dekker: Hi Mark, I just addressed your comment on `\r`. (Please check, as: [force-pushed](https://github.com/marzer/tomlplusplus/compare/61c233b50cae0826cf7724b2ace201e4f57c2725..81bff73982063fbc35698e5a17ee076af8bc1967)) Can you please take care of the ".inl" part?

marzer: Yep, sure.

marzer (approved):
🎉

include/toml++/impl/source_region.hpp
N-Dekker: Thanks Mark! By the way, I personally don't use `noexcept` so extensively, which is why I originally forgot to add one here. I became more reluctant to use it after seeing some disappointing performance effects from adding `noexcept`, from my [noexcept_benchmark](https://github.com/N-Dekker/noexcept_benchmark). Only for move constructors, move assignments, and `swap` functions, the benefit of `noexcept` looks obvious to me. But then again, marking get_line  `noexcept` is no problem to me, it won't harm 😃

marzer: I do not think of `noexcept` as a performance hint, but a documentation tool. If the function doesn't throw, I mark it as `noexcept` to indicate as much.

N-Dekker: Thanks Mark, I see your point. On the other hand, adding a `noexcept` specifier may also have the effect of causing a function to terminate, instead of throwing an exception. I think I often prefer an exception over an instant program termination. 🤷

N-Dekker:
Thank you very much for merging my feature request! Really cool!

Would it be an option for you to tag the master branch, including both of my contributions? In general, we prefer to use tagged versions of third party dependencies, for [elastix](https://github.com/SuperElastix/elastix)!

include/toml++/impl/source_region.hpp
N-Dekker: By the way, technically it appears possible to declare `get_line` constexpr. Personally I don't need that right now, but maybe it's still something for you to consider...?

N-Dekker: Thanks Mark, I generally agree that a function that may leak exceptions should not be marked `noexcept`. (Except for _maybe_ a move constructor or move assignment operator.) So honestly I feel a bit unsure about C++ Core Guidelines item [Use `noexcept` when exiting a function because of a `throw` is impossible or unacceptable](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-noexcept) Which says:

> _Example_
```
vector<double> munge(const vector<double>& v) noexcept
{
    vector<double> v2(v.size());
    // ... do something ...
}
```
> The noexcept here states that I am not willing or able to handle the situation where I cannot construct the local vector. That is, I consider memory exhaustion a serious design error (on par with hardware failures) so that I’m willing to crash the program if it happens.

In general, I would not mark a function that might create a non-empty `std::vector` noexcept, would you?

PR #257, opened 2025-02-26
specify `get_line` behavior on out of range versus empty line

N-Dekker:
**Commit text**

When the specified line is empty, the returned `string_view` will still have a valid (non-null) data pointer. When the line number is out of range, the data pointer will be null.

**What does this change do?**

It documents the return value more specifically, when the specified line number is out of range, and when the line is empty. Allowing users to distinguish between those two cases more easily.

 - It's another follow-up to pull request #255 !

**Pre-merge checklist**

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ UTF-8 decoding is performed using a state machine based on Bjoern Hoehrmann's '[
 -   **[@levicki](https://github.com/levicki)** - Helped design some new features
 -   **[@moorereason](https://github.com/moorereason)** - Reported a whole bunch of bugs
 -   **[@mosra](https://github.com/mosra)** - Created the awesome [m.css] used to generate the API docs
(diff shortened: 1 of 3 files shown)


marzer:
Hmmmn, that's not a particularly idiomatic usage of string views. Perhaps returning `optional<string_view>` might be more palatable?

N-Dekker:
Ah, I see, thanks. I did not realize before that the use of the null-ness of a string_view appears controversial. But indeed, I understand from  the discussion on [string_view::is_null()](https://groups.google.com/a/isocpp.org/g/std-proposals/c/8SBZGcdkW50) that such a use case is discouraged. Even though the standard explicitly specifies that the data pointer of a default-constructed string_view is null 🤷 

So I'm now considering to use `optional<string_view>`, as you suggest. To be continued...

N-Dekker:
- Superseded by pull request #258 commit 2ca7ac6c2cf1ed91aa1134dc663b6601d5e4c545 (which was just merged)

PR #258, opened 2025-02-27
let `toml::get_line` return an _optional_ string_view

N-Dekker:
**Commit message**

let `toml::get_line` return an _optional_ string_view

Allows users to detect more easily when the specified line number is out of range.

**What does this change do?**

Changes the return type of `get_line` from `string_view` to `optional<string_view>`.

 - Aims to supersede pull request #257, following Mark's suggestion at https://github.com/marzer/tomlplusplus/pull/257#issuecomment-2685814256

**Pre-merge checklist**

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/tests/user_feedback.cpp
+++ b/tests/user_feedback.cpp
@@ -455,12 +455,12 @@ b = []
 	SECTION("tomlplusplus/issues/254") // https://github.com/marzer/tomlplusplus/issues/254
 	{
 		// Check constexpr support.
-		static_assert(toml::get_line(""sv, 1) == std::string_view{});
+		static_assert(!toml::get_line(""sv, 1));
 		static_assert(toml::get_line("alpha = 1\nbeta = 2\n # last line # "sv, 1) == "alpha = 1"sv);
 
 		for (const toml::source_index line_num : { 0u, 1u, 2u })
 		{
-			CHECK(toml::get_line(""sv, line_num) == std::string_view{});
+			CHECK(!toml::get_line(""sv, line_num));
 		}
 
 		for (const auto input : {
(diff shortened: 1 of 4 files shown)


marzer:
Aye, this is much better methinks. A default-constructed std::string_view is well-defined, but using them as bool-like is rather unusual (and makes the code somewhat less ergnonomic because they don't have `operator bool()`).

Also with newer standards the `optional<>` lets people use this with the monadic interface functions out-of-the box 😄

N-Dekker:
Cool! Thanks again Mark!

tests/user_feedback.cpp
```diff
@@ -455,12 +455,12 @@ b = []
 	SECTION("tomlplusplus/issues/254") // https://github.com/marzer/tomlplusplus/issues/254
 	{
 		// Check constexpr support.
-		static_assert(toml::get_line(""sv, 1) == std::string_view{});
+		static_assert(!toml::get_line(""sv, 1));
```
N-Dekker, line 458: By the way, I originally wanted to use `std::nullopt`, both in the implementation (when out of range, `return std::nullopt`) and the unit test (above here, as in `static_assert(toml::get_line(""sv, 1) == std::nullopt)`, but I wasn't sure if it would be compatible with the `TOML_OPTIONAL_TYPE` option. What do you think?

marzer: Yeah, that's correct. The library allows people to customize the `optional<T>` in use, so hard-coding `std::nullopt` would break that (e.g. some may be using `tl::optional` or similar).

PR #259, opened 2025-02-28
let error_printer example print source line of parse error

N-Dekker:
**What does this change do?**

It aims to demonstrate a typical use case for `toml::get_line`: printing the source line of a parse error.

- Follow-up to pull request #258.

**Pre-merge checklist**

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/examples/error_printer.cpp
+++ b/examples/error_printer.cpp
@@ -104,6 +104,39 @@ namespace
 
 	inline constexpr auto divider =
 		"################################################################################"sv;
+
+	void print_string(std::string_view str)
+	{
+		for (char c : str)
+		{
+			if (c >= 0 && static_cast<std::size_t>(c) < std::size(toml::impl::control_char_escapes))
+			{
+				std::cout << toml::impl::control_char_escapes[static_cast<std::size_t>(c)];
+			}
+			else
+			{
+				if (c == '\\')
+				{
+					std::cout << '\\';
+				}
+				std::cout << c;
+			}
+		}
+	}
+
+	void print_parse_error(std::string_view doc, const toml::parse_error& err)
+	{
+		std::cout << err;
+
+		auto line_num = err.source().begin.line;
+
+		if (auto line = toml::get_line(doc, line_num))
+		{
+			std::cout << "\nLine "sv << line_num << ": "sv;
+			print_string(*line);
+		}
+		std::cout << "\n\n"sv;
+	}
 }
 
 int main()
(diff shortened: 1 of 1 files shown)


examples/error_printer.cpp
```diff
@@ -118,7 +175,16 @@ int main()
...
+			std::cout << err;
+
+			auto line_num = err.source().begin.line;
+
+			if (auto line = toml::get_line(str, line_num))
+			{
+				std::cout << "\nLine "sv << line_num << ": "sv;
+				print(*line);
+			}
+			std::cout << "\n\n"sv;
```
marzer, line 187: Ah, nice. Though, note you've only added this to the exception-based path - the exceptionless one lacks this feature. Ideally you'd split this new logic so that the get_line part would come after the exception branching, since it's the same in both cases.

examples/error_printer.cpp
```diff
@@ -104,6 +104,63 @@ namespace
...
+			"\\u0017"sv,
+			"\\u0018"sv,
+			"\\u0019"sv,
+			"\\u001A"sv,
+			"\\u001B"sv,
+			"\\u001C"sv,
+			"\\u001D"sv,
+			"\\u001E"sv,
+			"\\u001F"sv,
+		};
```
N-Dekker, line 145: This list was copied from forward_declarations.hpp, of course! @marzer Would it be OK if the example would directly use `impl::control_char_escapes` instead? I guess not, because it's an implementation detail. Just asking anyway  🤷  https://github.com/marzer/tomlplusplus/blob/2ca7ac6c2cf1ed91aa1134dc663b6601d5e4c545/include/toml%2B%2B/impl/forward_declarations.hpp#L127-L161

marzer: I'm not fussed to be honest. That particular example isn't really intended as a code example, but more of an example of how the error messages themselves are formatted.

N-Dekker: OK, so would you prefer "error_printer.cpp" to use `control_char_escapes` from "impl/forward_declarations.hpp"? Or you you mean it doesn't really matter to you?

marzer: I mean if you want to simplify it by re-using the implementation detail bit, rather than copying it, go ahead, I'm not going to be a purist about it.

N-Dekker: Hereby [force-pushed](https://github.com/marzer/tomlplusplus/compare/ffa0d3fecdd8106a6c0e7a1af1b224512d7fcee8..ac52b9e521d016992f21120e770268a2d7aaaa5d), please check!

Honestly I _am_ a bit of a purist on trying to avoid repetitive code, trying to follow the DRY principle 😇

examples/error_printer.cpp
N-Dekker: Just in case you missed it, I have addressed this comment of yours in the mean time, by adding a local helper function, `print_parse_error`, called by both the exception-based and the exceptionless path.

marzer: Yep, I saw, looks good :)

marzer (approved):
🎉

Issue #261, opened 2025-03-06
clang++-20 now warns about 'preceding whitespace in literal operator declaration deprecated'


eddelbuettel:
My [rcpptoml](https://github.com/eddelbuettel/rcpptoml) package uses `toml++` downstream with R, and the maintainers of the R repo test quite proactively.  As can be seen [here](https://cran.r-project.org/web/checks/check_results_RcppTOML.html), under Fedora with `clang++-20` we now get

```sh
Found the following significant warnings:
  ../inst/include/toml++/impl/path.h:793:19: warning: identifier '_tpath' preceded by whitespace in a literal operator declaration is deprecated [-Wdeprecated-literal-operator]
  ../inst/include/toml++/impl/parser.h:348:27: warning: identifier '_toml' preceded by whitespace in a literal operator declaration is deprecated [-Wdeprecated-literal-operator]
```

which I will of course fix in the package, simply be removing the whitespace. (Just did the same for a package using the `date` library.)  But as it is present in `toml++` I thought I should bring it up and offer a simple PR if that is desired.

A recursive search via `ag` shows (for the current HEAD):

```txt
-*- mode: ag; default-directory: "~/git/tomlplusplus/" -*-
Ag started at Thu Mar  6 13:54:20

ag --literal --group --line-number --column --color --color-match 30\;43 --color-path 1\;32 --smart-case --stats -- operator\"\"\  .
File: toml.hpp
3661:8:		path operator"" _tpath(const char* str, size_t len)
9715:16:		parse_result operator"" _toml(const char* str, size_t len)
9724:16:		parse_result operator"" _toml(const char8_t* str, size_t len)

File: tests/tests.hpp
48:18:constexpr size_t operator"" _sz(unsigned long long n) noexcept

File: include/toml++/impl/path.hpp
793:8:		path operator"" _tpath(const char* str, size_t len)

File: include/toml++/impl/parser.hpp
348:16:		parse_result operator"" _toml(const char* str, size_t len)
377:16:		parse_result operator"" _toml(const char8_t* str, size_t len)
7 matches
4 files contained matches
197 files searched
3777949 bytes searched
0.004609 seconds

Ag finished at Thu Mar  6 13:54:20
```

There is also this, but maybe less imminent as it is only in vendored code:

```
-*- mode: ag; default-directory: "~/git/tomlplusplus/" -*-
Ag started at Thu Mar  6 13:51:54

[...]

PR #263, opened 2025-03-11
Address clang++20 nag about 'whitespace in literal operator'

eddelbuettel:
As discussed in #261, `clang++-20` now nags about a whitespace after the `"""` in `operator""`.  A simple and sufficient change is to remove the whitespace. I have done so in `RcppTOM` and the CRAN repository for R which is already running some checks with `clang++-20` no longer tags the package.

I carried the same change over here.

Note that the embedded `catch` and `nlohmann::json` have the same issue in `operator "" ` (the initial whitespace between `operator` and `""` is apparently permitted) in a handful of places but as this is only used for testing I left it alone.

**Pre-merge checklist**

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/include/toml++/impl/parser.hpp
+++ b/include/toml++/impl/parser.hpp
@@ -345,7 +345,7 @@ TOML_NAMESPACE_START
 		///				A toml::parse_result.
 		TOML_NODISCARD
 		TOML_ALWAYS_INLINE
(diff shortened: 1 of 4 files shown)


marzer:
Thanks 🎉

PR #264, opened 2025-03-14
Fix build msys2

Toni500github:
**What does this change do?**
it fixes a compilation error from windows msys2, because looks like `impl::widen(file_path_str).c_str()` returns a `const wchar_t*` which seems to cause the errors.
I had a workaround to put that TOML_WINDOWS is false and it compiled, so I've decided to remove it.
image

no

**Pre-merge checklist**

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/include/toml++/impl/parser.inl
+++ b/include/toml++/impl/parser.inl
@@ -3784,11 +3784,7 @@ TOML_ANON_NAMESPACE_START
 		std::ifstream file;
 		TOML_OVERALIGNED char file_buffer[sizeof(void*) * 1024u];
 		file.rdbuf()->pubsetbuf(file_buffer, sizeof(file_buffer));
-#if TOML_WINDOWS
-		file.open(impl::widen(file_path_str).c_str(), std::ifstream::in | std::ifstream::binary | std::ifstream::ate);
-#else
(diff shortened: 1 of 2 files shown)


include/toml++/impl/parser.inl
```diff
@@ -3784,11 +3784,7 @@ TOML_ANON_NAMESPACE_START
 		std::ifstream file;
 		TOML_OVERALIGNED char file_buffer[sizeof(void*) * 1024u];
 		file.rdbuf()->pubsetbuf(file_buffer, sizeof(file_buffer));
-#if TOML_WINDOWS
```
marzer, line 3787: Erm... the solution to this problem is not to just remove this logic. That only fixes it for you, and does not take into consideration the use case that led to me implementing this way to begin with.

The solution to this problem is to detect (WINDOWS && !MSYS2).

marzer (changes requested):

include/toml++/impl/parser.inl
Toni500github: updated

marzer (approved):
Much better. Thanks 😄 🎉

Issue #172, opened 2022-09-13
Fail to build when imported from a module unit purview


wroyca:
*— This problem is reported here more as a backlog than as a real problem. It shouldn't really concern people who don't use state-of-the-art compilers with C++ modules. Nevertheless, @marzer asked me to flag it as a problem to look into at some point.* 

---

**Commit:**
**c6deadf61db048feda3998041af8021244368671**

**Compiler:**  
[Microsoft Visual C++](https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B) 14.32.31326 / May 10, 2022

**C++ standard mode:**  
23

**Target arch:**  
x64

[tomlplusplus](https://github.com/marzer/tomlplusplus) will flag [C2872](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c2872) should we try to use it when it is imported from a module unit purview: 

```cpp
import <sstream>;
import <iostream>;

export module uniform; // purview begin...
import "toml.hpp";     // import inside uniform's purview...

using namespace std::string_view_literals;

int main()
{
  static constexpr std::string_view some_toml = R"(
    [library]
    name = "toml++"
    authors = ["Mark Gillard <mark.gillard@outlook.com.au>"]
    cpp = 17
  )"sv;

  try
  {
    // parse directly from a string view:
    {
      toml::table tbl = toml::parse(some_toml);
      std::cout << tbl << "\n";
    }

    // parse from a string stream:
    {
      std::stringstream ss{ std::string{ some_toml } };
      toml::table tbl = toml::parse(ss); // Error! C2872!
      std::cout << tbl << "\n";
    }
  }
  catch (const toml::parse_error& err)
  {
    std::cerr << "Parsing failed:\n" << err << "\n";
    return 1;
  }

  return 0;
}
```

``` bash
> Error C2079 'toml::v3::impl::impl_ex::parser::root' uses undefined class 'toml::v3::table'
```

PR #266, opened 2025-03-28
Allow tomlplusplus to be built by C++20 modules

mikomikotaishi:
**What does this change do?**

This allows tomlplusplus to optionally be built with C++20 modules.

No, but this may potentially be a suggested feature by users of C++20 and later. (Mentioned by #172, though that is for a header unit rather than a module.)

**Pre-merge checklist**

    -   [x] Clang 8 or higher
    -   [x] GCC 8 or higher


mikomikotaishi:
Alternatively the module that it exports could be named just "tomlpp" instead of "tomlplusplus", for convenience

CMakeLists.txt
```diff
@@ -45,3 +45,10 @@ if(PROJECT_IS_TOP_LEVEL)
     add_subdirectory(fuzzing)
   endif()
 endif()
+
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.28)
+  message(STATUS "Building C++ modules (CMake ${CMAKE_VERSION} supports modules)")
+  add_subdirectory(modules)
+else()
+  message(STATUS "Skipping C++ modules (requires CMake 3.28+, found ${CMAKE_VERSION})")
+endif()
```
marzer, line 54: Ideally this should also be configurable with an opt-out cmake option, as I forsee this logic causing issues in some contexts. The tooling surrounding modules is quite new across the board, after all.

In fact, said option should probably be `OFF` by default, to maintain compatibility with current workflows. I'd hate to have exisiting users take this new version and suddenly find their build breaks because modules are now turned on.

marzer (commented):
Ooooh! Very exciting!

> Is it related to an exisiting bug report or feature request?
> No, but this may potentially be a suggested feature by users of C++20 and later.

There was an issue related to modules, #172, though the submitter closed it for some reason. (cc @wroyca)

CMakeLists.txt
mikomikotaishi: Done

wroyca:
Hello!


That was 3 years ago; I don't remember why I closed it. If anything, modules were in a much worse state back then, so it's really not a good reference point. Feel free to treat that old PR as if it didn't exist. :)

mikomikotaishi:
Hi, are there still any changes you are looking for before you accept and merge my pull request?

modules/tomlpp.cppm
```diff
@@ -0,0 +1,66 @@
...
+    using toml::source_index;
+    using toml::source_path_ptr;
+    using toml::table_iterator;
+
+    using toml::at_path;
+    using toml::operator""_toml;
+    using toml::operator""_tpath;
+    using toml::operator<<;
+    using toml::parse;
+    using toml::parse_file;
```
N-Dekker, line 65: Very interesting @mikomikotaishi

How did you select the symbols to be exported here, by these `using`'s? Could it be that `using toml::get_line` should still be added here? From https://github.com/marzer/tomlplusplus/blob/f5b3fb2c3e0090a05ea5ee82d0305e7bd95bacab/include/toml%2B%2B/impl/source_region.hpp#L241 (My contribution, honestly  😸 )

modules/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    tomlplusplus::tomlplusplus
+)
+
+target_include_directories(tomlplusplus_modules
+  PRIVATE
+    ${tomlplusplus_SOURCE_DIR}/include
+)
+
+set_target_properties(tomlplusplus_modules PROPERTIES 
+  CXX_STANDARD 23
```
N-Dekker, line 31: For my understanding, would `CXX_STANDARD 20` (rather than `23`) also be sufficient?

marzer:
@mikomikotaishi 

> Hi, are there still any changes you are looking for before you accept and merge my pull request?

Two things:

- Your files are missing trailing line endings. A minor annoyance, but please address it.
- @N-Dekker asks some good questions that I'm also curious to hear the answer to.

After that I'm happy to see this merged 👍🏼

modules/tomlpp.cppm
mikomikotaishi: I just exported everything I found in the documentation page, I could certainly add these now that I'm aware these symbols exist and are meant to be part of the public API

modules/CMakeLists.txt
mikomikotaishi: My apologies - C++20 is sufficient, I'll have that fixed

mikomikotaishi:
> @mikomikotaishi
> 
> > Hi, are there still any changes you are looking for before you accept and merge my pull request?
> 
> Two things:
> 
>     * Your files are missing trailing line endings. A minor annoyance, but please address it.
> 
>     * @N-Dekker asks some good questions that I'm also curious to hear the answer to.
> 
> 
> After that I'm happy to see this merged 👍🏼

OK, changes have been made as requested

N-Dekker:
@mikomikotaishi Thanks very much for addressing my questions regarding the CMake variable `CXX_STANDARD` and `toml::get_line`! How did you test your new feature? I'm wondering if there could be some kind of test for this feature on the CI as well, not sure if that's feasible 🤷  Ideally I guess all unit tests could optionally be built using the module, instead of the header file. As in:

```cpp
#if TOML_TEST_MODULE
import toml;
#else
#include <toml++/toml.hpp>
#endif

```
The tests could then be built either with `TOML_TEST_MODULE=0` or with `TOML_TEST_MODULE=1`, and both should then work equally well, right? I don't know if that's feasible for now. But at least _some_ kind of test of this interesting new feature would certainly be nice. What do you think?

mikomikotaishi:
> @mikomikotaishi Thanks very much for addressing my questions regarding the CMake variable `CXX_STANDARD` and `toml::get_line`! How did you test your new feature? I'm wondering if there could be some kind of test for this feature on the CI as well, not sure if that's feasible 🤷  Ideally I guess all unit tests could optionally be built using the module, instead of the header file. As in:
> 
> 
> 
> ```cpp
> 
> #if TOML_TEST_MODULE
> 
> import toml;
> 
> #else
> 
> #include <toml++/toml.hpp>
> 
> #endif
> 
> 
> 
> ```
> 
> The tests could then be built either with `TOML_TEST_MODULE=0` or with `TOML_TEST_MODULE=1`, and both should then work equally well, right? I don't know if that's feasible for now. But at least _some_ kind of test of this interesting new feature would certainly be nice. What do you think?

To be honest, I haven't written any example tests bundled with this project but I do have a basically identical copy (only difference is I aliased all type names to PascalCase) of all the files in this pull request in one of my own projects, which compiles and uses tomlplusplus fine.

N-Dekker:
Thanks @mikomikotaishi Honestly I do have quite some CMake experience, but I haven't yet tried out C++ modules, I have to admit. I'm still stuck at C++17 🤷 

When your pull request is merged, would you use it yourself, for [your project](https://github.com/mikomikotaishi/openJuice), replacing your original "openJuice/tomlpp/tomlpp.cppm" file?

src/modules/tomlpp.cppm
```diff
@@ -0,0 +1,67 @@
...
+ */
+export namespace toml {
+    /**
+     * @namespace literals
+     * @brief The toml++ namespace toml::literals::
+     */
+    inline namespace literals {
+        using toml::literals::operator""_toml;
+        using toml::literals::operator""_tpath;
+    }
```
N-Dekker, line 24: Do you use Visual C++? I just tried it out, using the latest update of Visual Studio 2022 (version 17.13.5). Unfortunately if fails to export the operators from `toml::literals`, saying:

```
>S:\tomlplusplus\src\modules\tomlpp.cppm(22,21): error C2872: 'literals': ambiguous symbol
>    S:\tomlplusplus\src\modules\tomlpp.cppm(21,22):
>    could be 'toml::literals'
>    S:\tomlplusplus\include\toml++\impl\parser.hpp(322,19):
>    or       'toml::v3::literals'
>S:\tomlplusplus\src\modules\tomlpp.cppm(22,31): error C2039: '""_toml': is not a member of 'toml::literals'
>    S:\tomlplusplus\src\modules\tomlpp.cppm(21,22):
>    see declaration of 'toml::literals'
>S:\tomlplusplus\src\modules\tomlpp.cppm(22,9): error C2873: '""_toml': symbol cannot be used in a using-declaration
>S:\tomlplusplus\src\modules\tomlpp.cppm(23,21): error C2872: 'literals': ambiguous symbol
>    S:\tomlplusplus\src\modules\tomlpp.cppm(21,22):
>    could be 'toml::literals'
>    S:\tomlplusplus\include\toml++\impl\parser.hpp(322,19):
>    or       'toml::v3::literals'
>S:\tomlplusplus\src\modules\tomlpp.cppm(23,31): error C2039: '""_tpath': is not a member of 'toml::literals'
>    S:\tomlplusplus\src\modules\tomlpp.cppm(21,22):
>    see declaration of 'toml::literals'
>S:\tomlplusplus\src\modules\tomlpp.cppm(23,9): error C2873: '""_tpath': symbol cannot be used in a using-declaration
```
Do you have a clue how to fix this?

N-Dekker:
I just tried your module in simple_parser.cpp. Did a CMake build, having `CMAKE_CXX_STANDARD "20"`, `TOMLPLUSPLUS_BUILD_MODULES` and `BUILD_EXAMPLES`. Added `import tomlplusplus;` as first line of code, and removed `#include <toml++/toml.hpp>`. Here: https://github.com/marzer/tomlplusplus/blob/f5b3fb2c3e0090a05ea5ee82d0305e7bd95bacab/examples/simple_parser.cpp#L9 Added the generated "tomlpp.cppm.ifc" file to Visual Studio, simple_parser, C/C++ Additional Module Dependencies.

Unfortunately the compiler says:
```
>------ Build started: Project: simple_parser, Configuration: Debug x64 ------
>simple_parser.cpp
>S:\tomlplusplus\include\toml++\impl\value.hpp(270,39): error C3878: syntax error: unexpected token ',' following 'simple-type-specifier'
>(compiling source file 'S:/T/tomlplusplus/examples/simple_parser.cpp')
>    S:\tomlplusplus\include\toml++\impl\value.hpp(270,39):
>    missing one of: '(' '{' ?
>    S:\tomlplusplus\include\toml++\impl\value.hpp(270,39):
>    the template instantiation context (the oldest one first) is
>        S:\tomlplusplus\include\toml++\impl\parser.inl(2757,26):
>        while compiling class template member function 'toml::v3::value<int64_t>::value<int64_t,0>(int64_t &&) noexcept(<expr>)'
>S:\tomlplusplus\include\toml++\impl\value.hpp(269,43): error C2057: expected constant expression
>(compiling source file 'S:/T/tomlplusplus/examples/simple_parser.cpp')
>Done building project "simple_parser.vcxproj" -- FAILED.

```
 
Not sure if it's an MSVC bug, I'm using the latest VS2022 update (version 17.13.5). Any clue?

src/modules/tomlpp.cppm
mikomikotaishi: That's very strange. I use Clang and no such error occurs on my end. I can look at it a bit more closely in a bit

mikomikotaishi:
> Thanks @mikomikotaishi Honestly I do have quite some CMake experience, but I haven't yet tried out C++ modules, I have to admit. I'm still stuck at C++17 🤷 
> 
> 
> 
> When your pull request is merged, would you use it yourself, for [your project](https://github.com/mikomikotaishi/openJuice), replacing your original "openJuice/tomlpp/tomlpp.cppm" file?

That file tomlpp.cppm exports all the symbols in PascalCase, which is the convention I am using in that project, so if I were to use it I would have to make a few adjustments to re-exporting them in my own project the way I aliased them there. At the moment I would probably just replace the #include with an import there, because modules can't be re-aliased/renamed

src/modules/tomlpp.cppm
N-Dekker: I don't know whether the "ambiguous symbol" errors are caused by a compiler bug. But it appears that these MSVC errors go away when explicitly using the operators from `toml::v3::literals`. As follows:

```cpp
    inline namespace literals {
        using toml::v3::literals::operator""_toml;
        using toml::v3::literals::operator""_tpath;
    }
```

Hope that helps!

mikomikotaishi: It may be an inline namespace thing I think, I'll try changing that to use the v3 scope and see if it works on Clang as well

mikomikotaishi:
> PS @mikomikotaishi Could you please check if simple_parser does compile (and run) with Clang, when you let it do import tomlplusplus; instead of #include <toml++/toml.hpp>?
Sure I'll test it and let you know, might take a bit of CMake trickery but given my past experience with it it does seem to work

mikomikotaishi:
@N-Dekker It seems to work on Clang for me, maybe it's a weird compiler issue

mikomikotaishi:
@marzer Any update on this, is it ready to be merged?

marzer:
Nope, looks all good, thanks 😄

gogoalalmh-dotcom (commented):
[``](url)

DmitriiKuchevskii:
Sorry for that, but I have a silly question
How do I use module? The only way I could make it work is
- Install headers into system dir
- Copy tomlplusplus.cppm into my project and add it into CXX_MODULES (in cmake file)

Is it somehow possible to import the module w/o adding .cppm file into the project?

mikomikotaishi:
Hi,

I am a little confused what you mean. The module is declared in that file, so naturally it must be part of the project to be imported. Could you clarify what it is you would like?

PR #267, opened 2025-03-31
fix MSVC C++20 module build error C2872, C2039, C2873 on 'literals'

N-Dekker:
**What does this change do?**

Replaced `toml` with `TOML_NAMESPACE` in the using-declarations in the exported namespace.

Fixed the errors from Visual Studio 2022 version 17.13.5 Compiler Version 19.43.34809 that appeared when building "tomlplusplus_modules", saying:

    src\modules\tomlpp.cppm(22,21): error C2872: 'literals': ambiguous symbol
        src\modules\tomlpp.cppm(21,22):
        could be 'toml::literals'
        include\toml++\impl\parser.hpp(322,19):
        or       'toml::v3::literals'
    src\modules\tomlpp.cppm(22,31): error C2039: '""_toml': is not a member of 'toml::literals'
        src\modules\tomlpp.cppm(21,22):
        see declaration of 'toml::literals'
    src\modules\tomlpp.cppm(22,9): error C2873: '""_toml': symbol cannot be used in a using-declaration

   - Follow-up to pull request #266 commit fea1d905f2d2a8ad830f1985fe879f4fd4601fe5 by @mikomikotaishi

**Pre-merge checklist**

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/src/modules/tomlpp.cppm
+++ b/src/modules/tomlpp.cppm
@@ -5,6 +5,7 @@
 
 module;
 
+#define TOML_UNDEF_MACROS 0
 #include <toml++/toml.hpp>
 
 export module tomlplusplus;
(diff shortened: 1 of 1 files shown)


src/modules/tomlpp.cppm
```diff
@@ -19,8 +20,8 @@ export namespace toml {
      * @brief The toml++ namespace toml::literals::
      */
     inline namespace literals {
-        using toml::literals::operator""_toml;
-        using toml::literals::operator""_tpath;
+        using TOML_NAMESPACE::literals::operator""_toml;
+        using TOML_NAMESPACE::literals::operator""_tpath;
     }
 
     using toml::array;
```
N-Dekker, line 27: @marzer Do you think

src/modules/tomlpp.cppm
```diff
@@ -19,8 +20,8 @@ export namespace toml {
      * @brief The toml++ namespace toml::literals::
      */
     inline namespace literals {
-        using toml::literals::operator""_toml;
-        using toml::literals::operator""_tpath;
```
N-Dekker, line 23: @marzer @mikomikotaishi I noticed that `operator""_toml` and `operator""_tpath` are exported twice. 

Here, inside `toml::literals`: https://github.com/marzer/tomlplusplus/blob/fea1d905f2d2a8ad830f1985fe879f4fd4601fe5/src/modules/tomlpp.cppm#L21-L24
And here, at the root of the `toml` namespace: https://github.com/marzer/tomlplusplus/blob/fea1d905f2d2a8ad830f1985fe879f4fd4601fe5/src/modules/tomlpp.cppm#L62-L63 Was that intended?

src/modules/tomlpp.cppm
marzer: Oh, hmmn, yeah. I completely forgot about all of that machinery.

In theory you should be able to simplify it by replacing 

```cpp
export namespace toml
{
	//...
}
```

with

```cpp
export TOML_NAMESPACE_START
{
	// ...
}
```

because `TOML_NAMESPACE_START` is shorthand for `namespace toml { inline namespace <internal abi namespace>`

src/modules/tomlpp.cppm
N-Dekker: @mikomikotaishi Well, it just isn't clear to me why `operator""_toml` and `operator""_tpath` are exported twice (once in `toml::literals` and once in `toml`), but if you have a use case that only works when they are exported twice, can you please share it?

Anyway, this pull request already appears to fix those MSVC compile errors on `literals`, so technically it seems to work, exporting those operators twice.

marzer:
@N-Dekker sorry, just coming back to this now. It looks complete, but was there anything still outstanding?

PR #268, opened 2025-04-01
avoid using `impl` in error_printer example

N-Dekker:
**What does this change do?**

Replaced access to `toml::impl::control_char_escapes` with manual handling of control characters, in `print_string`.

Would ease a possible future replacement of `#include <toml++/toml.hpp>` with `import tomplplusplus` (because `toml::impl` is not exported).

  -  It's another follow-up to pull request #259.

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher


examples/error_printer.cpp
```diff
@@ -110,23 +113,29 @@ namespace
...
 		{
-			if (c >= 0 && static_cast<std::size_t>(c) < std::size(toml::impl::control_char_escapes))
+			if (std::optional<std::string_view> escape_sequence = (c == '\\') ? "\\\\"sv
+																: (c == '\a') ? "\\a"sv
+																: (c == '\b') ? "\\b"sv
+																: (c == '\f') ? "\\f"sv
+																: (c == '\n') ? "\\n"sv
+																: (c == '\r') ? "\\r"sv
+																: (c == '\t') ? "\\t"sv
+																: (c == '\v') ? "\\v"sv
```
N-Dekker, line 123: Note: Two simple escape sequences, `\a` (0x07) and `\v` (0x0B), are not in `toml::impl::control_char_escapes`: https://github.com/marzer/tomlplusplus/blob/fea1d905f2d2a8ad830f1985fe879f4fd4601fe5/include/toml%2B%2B/impl/forward_declarations.hpp#L127-L142

I found them at https://eel.is/c++draft/lex.ccon#nt:simple-escape-sequence-char saying:

```
simple-escape-sequence-char: one of
  '  "  ?  \ a  b  f  n  r  t  v
```
Would it be worth adding these two to `toml::impl::control_char_escapes` as well? Honestly it doesn't matter much to me. Just asking 😃

marzer: They are in there:

- `\u0007`
- `\u000B`

This list is not used for parsing, so does not need to exhaustively list all possible representations. It's only used during serialization. Each entry in the array is at an index that corresponds to the equivalent codepoint.

N-Dekker: The `print_string` function in error_printer example is meant to print a human-readable output, so I guess for `print_string`, the human-friendly simple-escape-sequences `\a` and `\v` may be preferred, whereas for `impl::control_char_escapes`, `\u0007` and `\u000B` may be preferred. Right?

marzer: I mean it doesn't matter, either should work when writing TOML. But you can't add additional values to that impl array, add you suggested, because the index of the elements matches their content.

N-Dekker: OK, thanks! By the way, I noticed something weird (?) in the draft of the C++ Standard. [Table 10 - Simple escape sequences [tab:lex.ccon.esc]](https://eel.is/c++draft/tab:lex.ccon.esc) mentions both `\a` and `\v`. But there's another table, [Table 112 - Mapping of characters to escape sequences [tab:format.escape.sequences]](https://eel.is/c++draft/tab:format.escape.sequences), that doesn't mention them at all! Could it be that they accidentally forgot to mention those two, in table 112? It is part of a description of how to construct an escaped string, in section [[format.string.escaped]](https://eel.is/c++draft/format.string.escaped)

PR #271, opened 2025-05-07
TOML_DISABLE_NOEXCEPT_NOEXCEPT workaround MSVC bug noexcept + modules

N-Dekker:
**What does this change do?**

Allowed users of the tomlplusplus module to compile using MSVC.

Without `TOML_DISABLE_NOEXCEPT_NOEXCEPT`, MSVC appears to produce the following error:

> include\toml++\impl\value.hpp(270,39): error C3878: syntax error: unexpected token ',' following 'simple-type-specifier'
>    missing one of: '(' '{' ?
>    include\toml++\impl\value.hpp(270,39):
>    the template instantiation context (the oldest one first) is
>        include\toml++\impl\parser.inl(2757,26):
>        while compiling class template member function 'toml::v3::value<int64_t>::value<int64_t,0>(int64_t &&) noexcept(<expr>)'

- It is another follow-up to pull request #266

**Pre-merge checklist**

    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/include/toml++/impl/value.hpp
+++ b/include/toml++/impl/value.hpp
@@ -266,8 +266,11 @@ TOML_NAMESPACE_START
 			(impl::value_variadic_ctor_allowed<value<ValueType>, impl::remove_cvref<Args>...>::value),
 			typename... Args)
 		TOML_NODISCARD_CTOR
-		explicit value(Args&&... args) noexcept(noexcept(value_type(
-			impl::native_value_maker<value_type, std::decay_t<Args>...>::make(static_cast<Args&&>(args)...))))
+		explicit value(Args&&... args)
+#if !TOML_DISABLE_NOEXCEPT_NOEXCEPT
+			noexcept(noexcept(value_type(
+				impl::native_value_maker<value_type, std::decay_t<Args>...>::make(static_cast<Args&&>(args)...))))
+#endif
 			: val_(impl::native_value_maker<value_type, std::decay_t<Args>...>::make(static_cast<Args&&>(args)...))
 		{
 #if TOML_LIFETIME_HOOKS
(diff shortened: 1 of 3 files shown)


include/toml++/impl/value.hpp
```diff
@@ -266,8 +266,11 @@ TOML_NAMESPACE_START
 			(impl::value_variadic_ctor_allowed<value<ValueType>, impl::remove_cvref<Args>...>::value),
 			typename... Args)
 		TOML_NODISCARD_CTOR
-		explicit value(Args&&... args) noexcept(noexcept(value_type(
-			impl::native_value_maker<value_type, std::decay_t<Args>...>::make(static_cast<Args&&>(args)...))))
+		explicit value(Args&&... args)
+#if !TOML_DISABLE_NOEXCEPT_NOEXCEPT
+			noexcept(noexcept(value_type(
+				impl::native_value_maker<value_type, std::decay_t<Args>...>::make(static_cast<Args&&>(args)...))))
+#endif
```
N-Dekker, line 273: This is the only use of `noexcept(noexcept(<expression>))` in toml++, and while a regular MSVC build appears to be able to parse it, it fails when using the tomlplusplus module (doing a C++20 compilation).

N-Dekker: Honestly, MSVC _does_ compile `noexcept(noexcept(<expression>))` in other situations. I don't know exactly in which particular situations `noexcept(noexcept(<expression>))` triggers an error, but this is one of them  🤷

N-Dekker: Still "Draft" because I haven't yet figured out if this is a known compiler bug. Otherwise, maybe it should be reported!

The added value of this particular `noexcept` specifier looks rather small, right? 🤷  So then, is it OK to put it inside an `#if` ... `#endif`?

marzer: Hmmn, might be able to work around this by extracting the inner `noexcept()` out into a separate typedef or template variable, then using _that_ in the outer `noexcept()`?

Otherwise yeah I'm OK with a switch to turn it off.

N-Dekker: My feeling is that the combination of variadic templates + noexcept(noexcept(...)) + modules is more than MSVC can handle. You can use each feature, just don't combine them  😸

PR #273, opened 2025-05-09
Export missing symbols and fix CMakeLists.txt for modules

mikomikotaishi:
**What does this change do?**
This commit amends the module file to export some missing symbols, such as those of the name `is_*`, updates `CMakeLists.txt` in `modules` to fix a failure to compile, and renames `tomlpp.cppm` to `tomlplusplus.cppm`, so that the file name aligns with the name of the module. 

No, but it is related to a previous commit (#266).

**Pre-merge checklist**

    -   [x] Clang 8 or higher
    -   [x] GCC 8 or higher


mikomikotaishi:
@marzer @N-Dekker I noticed some parts of the `CMakeLists.txt` were removed (the `target_sources()` call, unless I'm mistaken), I ended up re-adding it because it wouldn't compile without it, could you remind me if there was a reason it was removed? Making these modifications allows me to use all the symbols perfectly fine though, and so I'm hoping to make sure there isn't something I'm missing on my end.

mikomikotaishi:
Are there any concerns or questions about my PR before you can have it merged?

marzer:
Hi there, erm, no concerns for me personally. I'm only keeping an eye on the modules stuff from a distance, so any changes here I think are largely between you and @N-Dekker to figure out 😅 (I don't know what the reasoning was behind the symbol export changes 🤷🏼‍♂️)

mikomikotaishi:
OK, sounds good. Let's hear what @N-Dekker has to say about the CMake changes then.

N-Dekker:
@marzer Thanks for merging my pull request #271 Sorry I kept it in "draft" status for too long! Obviously it's just a workaround, to make MSVC happy!

@mikomikotaishi Can you please rebase you pull request branch, so that I can try it out with MSVC?

Check-out your branch ([patch-2](https://github.com/mikomikotaishi/tomlplusplus/tree/patch-2)), and do:

    git fetch upstream master:master
    git rebase master
    git push --force

Assuming `upstream` is the name you have for the remote https://github.com/marzer/tomlplusplus.git

N-Dekker:
> Did you have any trouble getting it to work?

No, I just didn't have time yet. Is there a specific deadline for this feature, that I can take into account?

mikomikotaishi:
No, I suppose not, but it would be nice to merge it as soon as possible. I have started using the tomlplusplus module myself and because it is missing those symbols it creates problems which I have to manually resolve.

N-Dekker:
As there are no unit tests for the module, I'm just locally adjusting an existing test (example), ["examples/error_printer.cpp"](https://github.com/marzer/tomlplusplus/blob/master/examples/error_printer.cpp), trying to get it to work with the module: Instead of the `#include <toml++/toml.hpp>`, I do:

    import tomlplusplus;
    #define TOML_EXCEPTIONS 1

Unfortunately it still produces link errors with MSVC

```
>error_printer.obj : error LNK2001: unresolved external symbol "public: static unsigned char const * const toml::v3::impl::utf8_decoder::state_table" (?state_table@utf8_decoder@impl@v3@toml@@2QBEB)
>error_printer.obj : error LNK2001: unresolved external symbol "private: static unsigned __int64 const toml::v3::impl::impl_ex::parser::max_nested_values" (?max_nested_values@parser@impl_ex@impl@v3@toml@@0_KB)
  ...
>error_printer.obj : error LNK2001: unresolved external symbol "public: static class std::basic_string_view<char,struct std::char_traits<char> > const toml::v3::impl::parse_integer_traits<10>::full_prefix" (?full_prefix@?$parse_integer_traits@$09@impl@v3@toml@@2V?$basic_string_view@DU?$char_traits@D@std@@@std@@B)
>B:\tomlplusplus\examples\Debug\error_printer.exe : fatal error LNK1120: 18 unresolved externals

``` 

Are you able to locally adjust your "examples/error_printer.cpp" to do  `import tomlplusplus`, and then build and run it?

mikomikotaishi:
I used it [here](https://github.com/mikomikotaishi/openJuice/blob/main/deps/tomlpp/tomlpp.cppm) in this file that just re-exports all the symbols in the library, and it compiles fine on Clang.

N-Dekker:
> I used it [here](https://github.com/mikomikotaishi/openJuice/blob/main/deps/tomlpp/tomlpp.cppm) in this file that just re-exports all the symbols in the library, and it compiles fine on Clang.

@mikomikotaishi Thank you for trying! But sorry, I feel it's a pity that you do not use the module implementation from toml++ itself. It's your own proposal! It will give other users more trust in this great new toml++ feature, when the original author of the feature uses it directly! 

> Also, can you explain why you added `#define TOML_EXCEPTIONS 1`? I am not sure what that does, as modules do not export macros.

Thanks for asking, @mikomikotaishi ! The problem is that the example itself does `#if TOML_EXCEPTIONS`: https://github.com/marzer/tomlplusplus/blob/8eb4012353d80696e624c0685cbd0baa69772a46/examples/error_printer.cpp#L164

I haven't yet figures out how to deal with macro's that are part of the interface, when using modules. If I understand correctly, end-users should be able to have such an `#if TOML_EXCEPTIONS` in their code, right, @marzer ? So then it should be exported somehow 🤷

mikomikotaishi:
@marzer @N-Dekker I see, however the matter with `TOML_EXCEPTIONS` deciding whether the implementations will throw an exception would have to be decided when `tomlplusplus` is compiled, not during the compilation of the translation units that consume the module. Thus it would not be up to those units to do so, but rather the user, who would have to specify that macro at compilation. So adding `TOML_EXCEPTIONS` inside the importing files probably won't do anything. 

> But yeah probably whatever that value is, be it defaulted or user-specified, should find it's way into the code that consumes the module.

Modules cannot export macros, header units (headers that are `import`ed rather than `#include`ed) can however.

marzer:
> Modules cannot export macros, header units (headers that are imported rather than #includeed) can however.

I mean, sure, but that's somewhat a distinction without a difference, and not really what I meant. Header units are bridging measure introduced as part of the broader C++ modules language feature. People consuming the library that way need the macro in the same way people using regular `#includes` do.

> Anyway, could the PR be merged now?

Yeah, sure, it looks fine to me

PR #280, opened 2025-08-19
Fix missing include

SGauvin:
date_time.hpp was not including std_optional.hpp directly. It probably got it transitively from other files including it directly. This caused clangd to complain at every `#include <toml++/toml.hpp>`

**What does this change do?**

Adds a missing include

No

**Pre-merge checklist**

-   [x ] I've read [CONTRIBUTING.md]
-   [x ] I've rebased my changes against the current HEAD of `origin/master` (if necessary)
    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/include/toml++/impl/date_time.hpp
+++ b/include/toml++/impl/date_time.hpp
@@ -7,6 +7,7 @@
 #include "forward_declarations.hpp"
 #include "print_to_stream.hpp"
 #include "header_start.hpp"
+#include "std_optional.hpp"
 
 TOML_NAMESPACE_START
 {


include/toml++/impl/date_time.hpp
```diff
@@ -7,6 +7,7 @@
 #include "forward_declarations.hpp"
 #include "print_to_stream.hpp"
 #include "header_start.hpp"
+#include "std_optional.hpp"
```
marzer, line 10: Order matters here; header_start.hpp needs to be last.

Issue #28, opened 2020-05-20
Preservation of comments and whitespace


ned14:
Firstly, this is one of those very few libraries I've used where I don't think I'd have done much different myself. Well, possibly I'd have made it always non-throwing. Well done on that, and thank you for making it open source.

Another thing that I would have done is optional comment and whitespace preservation, such that if you read in a TOML file, and then write it out, comments and whitespace are preserved. This is useful because I'm using your library a bit like how clang-tidy or doxygen do things, where you can dump out a default TOML config, hand edit it, or supply TOML delta files to merge against the default TOML config, and so on. Therefore, dumping out the post-merge TOML file, with all comments also merged, would be ideal.

For this to work right, you'd need to retain comments and whitespace as an attribute of the first node following the comment. Then, if the node gets replaced by a merge, one could retain the existing comment and whitespace, or replace it with the new comment and whitespace etc.

I'd personally make this opt-in, as plenty of people don't want to waste CPU nor memory on comments and whitespace. My thanks in advance to you for your consideration!

marzer:
> Firstly, this is one of those very few libraries I've used where I don't think I'd have done much different myself. 

Damn, that's high praise. Thanks! Glad you like it. It's the first general-purpose C++ library I've released for public consumption so it's been nice to see it well-received.

> Another thing that I would have done is optional comment and whitespace preservation, such that if you read in a TOML file, and then write it out, comments and whitespace are preserved. This is useful because I'm using your library a bit like how clang-tidy or doxygen do things, where you can dump out a default TOML config, hand edit it, or supply TOML delta files to merge against the default TOML config, and so on. Therefore, dumping out the post-merge TOML file, with all comments also merged, would be ideal.

Yeah, this is a piece of functionality I've been asked about a couple of times now, though you're the first to actually provide a compelling use-case. It's tentatively on my mental road-map for `toml++` though might be a decent while before I can make time to implement it.

marzer:
Also, reciprocal praise: your work on profiling the STL headers (e.g. [here](https://www.reddit.com/r/cpp/comments/860hya/stl_header_token_parsing_benchmarks_for_vs2008/)) is something I referred to on a regular basis in optimizing compile times for the library

ned14:
You're very kind. I think if you're a certain kind of C++ dev, stuff like not including headers and, generally speaking, not having the compiler generate opcodes in the first place, leads to a certain shared philosophy. One which I would prefer to see more of on WG21, but that's by the by.

[...]

PR #283, opened 2025-09-18
Trivia preservation

stackotter:
**What does this change do?**

Implements optional table order and trivia preservation.

I have included table order preservation in this PR because fine-grained trivia preservation doesn't make much sense if large blocks of the document get reordered regardless. I can split table order preservation into a separate PR if you'd like.

*This is my first time writing CPP (although I've written plenty of C in the past), so just let me know if there's anything I've got wrong. I think I'm starting to understand move semantics and stuff like that, but I made quite a few mistakes earlier on (some of which led to crashes), and I probably haven't caught all of them.*

Addresses #28 (and extends its goals to include all trivia)

**Todo**

- [x] Implement table order preservation (gated behind `-Dordered_tables=true`)
- [x] Implement trivia preservation in parser (gated behind a runtime `collect_trivia` parameter available on all top-level parsing functions)
- [x] Implement trivia recreation in toml formatter (gated behind `format_flags::preserve_source_trivia`)
- [x] Support whitespace and comments
- [ ] Preserve numeric literal encoding (e.g. `0x15` should remain hexadecimal after round-tripping)
- [ ] Preserve string encoding (e.g. `'foo'` shouldn't become '"foo"' after round-tripping)
- [ ] Test against a wide variety of TOML documents to find syntax that fails to be successfully round-tripped
- [ ] Sensibly infer trivia when inserting toml values into an existing document with preserved trivia.

**Trivia inference**

Trivia inference for inserted toml values won't be perfect of course, but should at least be able to respect when e.g. a document only has single quotes, or always uses 4 space indentation. I'm thinking perhaps trivia preferences could be 'dragged' down through the document. So e.g. a new array element will use the style of the previous array element (if any). We could possibly start the 'drag' operation below the new element before wrapping around to the top in order to syntax that comes later in the document if there aren't any examples before the inserted syntax. This approach sounds slow describing it like that, but we'd probably be able to cache the dragged preferences at each node and only update them when inserting new elements.

**Pre-merge checklist**

-   [ ] I've rebased my changes against the current HEAD of `origin/master` (if necessary)
    -   [ ] Clang 8 or higher
    -   [ ] GCC 8 or higher

### DIFF
--- a/include/toml++/impl/array.hpp
+++ b/include/toml++/impl/array.hpp
@@ -7,6 +7,8 @@
 #include "std_utility.hpp"
 #include "std_vector.hpp"
 #include "std_initializer_list.hpp"
+#include "preprocessor.hpp"
+#include "trivia_piece.hpp"
 #include "value.hpp"
 #include "make_node.hpp"
 #include "header_start.hpp"
@@ -290,6 +292,7 @@ TOML_NAMESPACE_START
 		using vector_iterator		= typename vector_type::iterator;
 		using const_vector_iterator = typename vector_type::const_iterator;
 		vector_type elems_;
+		optional<std::vector<trivia_piece>> inner_trailing_trivia_;
 
 		TOML_NODISCARD_CTOR
 		TOML_EXPORTED_MEMBER_FUNCTION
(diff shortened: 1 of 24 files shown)


marzer:
Hmmn, there's a lot going on here! First note would be: you're doing this work against a commit that's 40 commits behind the current `main`, so there's likely to be a lot of old code and merge issues. I'd recommend addressing that before you go any further 😅

marzer:
Also, another note: This will introduce a lot of overhead in the storage, so please make this entire feature opt-in at compile-time. Every single piece of it should be gated behind some sort of `#if TOML_ENABLE_BLAH`, as otherwise the overheads are likely to be unacceptable for existing users.

marzer:
Oh and the trivia inference thing sounds like a bad idea. Please don't implement that.

PR #287, opened 2025-12-23
Add unit test build to OSS-Fuzz build.sh

arthurscchan:
OSS-Fuzz has a new Chronos feature that allows fast rebuilding of OSS-Fuzz-integrated projects and checks whether the project code or fuzzers still pass their unit tests after patching. To enable this feature, the unit tests of integrated projects need to be built. Therefore, this PR modifies the build script used by OSS-Fuzz to also build the unit tests for this project. The OSS-Fuzz-side changes for this project's integration can be found here: https://github.com/google/oss-fuzz/pull/14639.

### DIFF
--- a/fuzzing/build.sh
+++ b/fuzzing/build.sh
@@ -6,3 +6,38 @@ cmake -S . -B build -DBUILD_FUZZER=ON && cmake --build build --target install
 mkdir -p corpus
 find $SRC/tomlplusplus -name "*.toml" -exec cp {} corpus \;
 zip -q -j $OUT/toml_fuzzer_seed_corpus.zip corpus/*
+
+# Build unit test
+clang++ -std=c++17 -O2 -DUSE_VENDORED_LIBS=1 \
+    -Iinclude -Itests \
+    tests/at_path.cpp \
+    tests/conformance_burntsushi_invalid.cpp \
+    tests/conformance_burntsushi_valid.cpp \
+    tests/conformance_iarna_invalid.cpp \
+    tests/conformance_iarna_valid.cpp \
+    tests/formatters.cpp \
+    tests/for_each.cpp \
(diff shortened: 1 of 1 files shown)


fuzzing/build.sh
```diff
@@ -6,3 +6,38 @@ cmake -S . -B build -DBUILD_FUZZER=ON && cmake --build build --target install
 mkdir -p corpus
 find $SRC/tomlplusplus -name "*.toml" -exec cp {} corpus \;
 zip -q -j $OUT/toml_fuzzer_seed_corpus.zip corpus/*
+
+# Build unit test
+clang++ -std=c++17 -O2 -DUSE_VENDORED_LIBS=1 \
```
marzer, line 11: Explicitly invoking the compiler seems unnecessary here, given CMake is available and used above. Can't you just turn tests on in the CMake configure on line 3?

arthurscchan: Hi, thank you for your comments. I have reviewed the CMake configuration in the root CMakeLists.txt, but found that it does not enable or define any unit testing settings for this project. I did notice that the meson.build file in the tests directory defines unit testing. However, this requires additional meson and ninja-build tools to be installed in the OSS-Fuzz image, which is not ideal. Therefore, I examined how the unit testing is set up in meson.build and used only the compiler to build those unit tests.
Please do suggest if I am missing something here. Thanks.

marzer: Oh, sorry, nope you're not missing anything. I didn't realize the CMake integration didn't also build the tests (that integration was provided by a contributor, been a while since I looked at it).

No worries, this is fine as-is. I guess at some point I'll have to do a pass over the build systems and normalize them to fix this inconsistency.

marzer:
Thanks :)

PR #302, opened 2026-05-11
fix(parser): preserve quoted-key text in redefinition error messages

### LINKED ISSUE #300: wrong string in error message
## Environment
**toml++ version and/or commit hash:**  

found in 3.4.0, still exists in e79458781e821720ff362a432bf9ae5b20cc95bd

**Compiler:**  

gcc 14.3.0

**C++ standard mode:**  

default (I think it doesn't matter for this bug)

**Target arch:**  

x86_64

**Library configuration overrides:**  

`TOML_EXCEPTIONS=0`

**Relevant compilation flags:**  

none

## Describe the bug

We use tomlplusplus in ckmame for reading config files. By accident I added two sections for the same name (let's call it "foo"), which triggered a parse error:
```
Parse error: Error while parsing table header: cannot redefine existing table '"fofoo"'
```
Note that it does not complain about "foo" but "fofoo".

## Steps to reproduce (or a small repro code sample)

This standalone demo reproduces the problem for me:
```
#include <iostream>
#include <string_view>
#define TOML_EXCEPTIONS 0
#include "toml.hpp"

int main()
{
    std::string_view input =
        "[\"foo\"]\n"
        "dat-directories = \"bar\"\n"
        "[\"foo\"]\n"           // duplicate - to trigger the error
        "dat-directories = \"baz\"\n";

    auto result = toml::parse(input);

    if (!result)
    {
        std::string msg(result.error().description());
        std::cout << "Parse error: " << msg << "\n\n";

        bool ok = msg.find("\"foo\"") != std::string::npos;
        std::cout << (ok ? "PASS" : "FAIL")
                  << ": key name in error message is "
                  << (ok ? "correct (\"foo\")" : "WRONG (expected \"foo\")")
                  << "\n";
        return ok ? 0 : 1;
    }

    std::cout << "FAIL: expected a parse error but got none\n";
    return 1;
}
```
To compile:
```
c++ -o toml_bug_demo toml_bug_demo.cc
```
running it gives:
```
Parse error: Error while parsing table header: cannot redefine existing table '"fofoo"'

FAIL: key name in error message is WRONG (expected "foo")
```
but should look like this:
```
Parse error: Error while parsing table header: cannot redefine existing table '"foo"'

PASS: key name in error message is correct ("foo")
```

SAY-5:
**What does this change do?**

Fixes the corrupted quoted-key string that appears in redefinition error messages. A key written as `["foo"]` previously produced an error of the form `cannot redefine existing table '"fofoo"'`. With this change the diagnostic shows `"foo"`, as written in the source.

Fixes #300.

**Pre-merge checklist**

    - [ ] Clang 8 or higher
    - [x] GCC 8 or higher

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,7 @@ UTF-8 decoding is performed using a state machine based on Bjoern Hoehrmann's '[
 -   **[@rafal-c](https://github.com/rafal-c)** - Added a formatting flag
 -   **[@jfsimoneau](https://github.com/jfsimoneau)** - Fixed a meson option
 -   **[@ShereeMorphett](https://github.com/ShereeMorphett)** - Fixed reinterpret_cast usage
+-   **[@SAY-5](https://github.com/SAY-5)** - Fixed a parser bug
 <br>
 
 ## Contact
--- a/include/toml++/impl/parser.inl
+++ b/include/toml++/impl/parser.inl
@@ -1686,10 +1686,7 @@ TOML_IMPL_NAMESPACE_START
 			TOML_ASSERT_ASSUME(is_string_delimiter(*cp));
 			push_parse_scope("string"sv);
 
-			// snapshot the recording buffer so we can rewind it alongside the input
-			// stream if the upcoming lookahead turns out to be a non-multi-line string
-			// (see go_back(2u) below). without this, the two lookahead code points end
-			// up recorded twice and pollute diagnostic messages (issue #300).
+			// snapshot length so the recording buffer can be rewound alongside go_back(2u) below
 			const auto recording_buffer_rollback_size = recording_buffer.length();
 
 			// get the first three characters to determine the string type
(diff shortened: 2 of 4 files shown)


marzer (commented):
This fix looks very LLM-coded. The way the comments are written, and the fact that my PR template was annihilated wholesale, yet it's requirements were still referenced.

... Am I right? If so, please remove the LLM junk comments, restore the PR description, and complete the template by hand.

SAY-5:
@marzer apologies for the noise earlier. Reopened with three things addressed:

1. Trimmed the two LLM-style narrative comments in `parser.inl` down to a single line near the snapshot, since the variable name and the surrounding `go_back(2u)` already explain the rollback.
2. Trimmed the matching comments in the regenerated `toml.hpp`.
3. Restored the PR description so it follows the template verbatim and ticked the boxes that actually apply.

The fix itself is unchanged. The PR currently lacks (a) updating README.md contributors and (b) verifying on Clang/MSVC. Happy to add the README entry on request, and to add those compiler check-marks once I can verify on those toolchains.

marzer (approved):

PR #303, opened 2026-05-19
Remove duplicate CMake in modules implementation

davidstone:
It looks like two different approaches to the same problem got committed to this file. Remove the more complicated approach and keep the simpler approach. This fixes CMake warnings about the same source file appearing multiple times.

**Pre-merge checklist**

-   [X] I've read [CONTRIBUTING.md]
-   [X] I've rebased my changes against the current HEAD of `origin/master` (if necessary)
-   [X] I've rebuilt and run the tests with at least one of:
    -   [X] Clang 8 or higher
    -   [ ] GCC 8 or higher
-   [X] I've added my name to the list of contributors in [README.md](https://github.com/marzer/tomlplusplus/blob/master/README.md)


marzer:
Great, thanks 🎉

Issue #175, opened 2022-10-20
Is it possible to add support for custom types?


CJCombrink:
**Is your feature request related to a problem? Please describe.**

Is it possible to add TOML support for custom types, example:
```
struct Point {
    int32_t x = 10;
    int32_t y = 20;
}
```

Which would probably look like the following in TOML
```
[custom_point]
x = 10
y = 20
```

**Describe the solution you'd like**

I would like to use it as follow:
```
const Point my_point = tbl["custom_point"].value<Point>();
```

**Additional context**

I am not yet using the library but working thought the API docs I don't see any mention of such options.

marzer:
Nope, there's no way to do that currently. It has been vaguely on my to-do list for a while, but I'm unlikely to get time to add the plumbing for this any time soon.

CJCombrink:
Thanks for the pointer. I have looked at toml11.
But, I like the API of this library a lot more.
I can probably get around this "limitation" for the sake of the cleaner API and then hope in future I can throw away boilerplate code when this gets added. 

I am used to [JSON for Modern C++](https://github.com/nlohmann/json) and this interface is quite familiar. 
Perhaps there are some lessens to be learned from their support for [Custom data source](https://github.com/nlohmann/json#custom-data-source)s in my code or this library. 
(I am guessing you are familiar with this nlohmann-json)

marzer:
> (I am guessing you are familiar with this nlohmann-json)

Yup, the way that library handles this particular bit of functionality is very nice. One thing to be mindful of though is that TOML is chiefly a config language, whereas JSON is also aimed at serialization, so being able to convert between TOML and structs of arbitrary complexity isn't stricly a design goal I'm ever going to work towards.

Having said that, I would still like to make it easier for obvious common cases like points, vectors et cetera - I've had to come up with my own workarounds for that in my own code that uses this library 😅

bobfang1992:
I wonder how this is possible without proper reflection support in C++.

marzer:
It's not in the general case; it always boils down to manually-written operator overloads or ADL tricks. You can use template machinery (e.g. SFINAE, partial specialisation) to cover some predefined cases pretty easily, though.

PR #306, opened 2026-07-26
Add opt-in support for custom string types

mikomikotaishi:
**What does this change do?**

This PR adds support for custom string types through an opt-in `toml::string_like<T>` specialisation (inherit `std::true_type` to enable), as long as it can be convertible from `std::string`/`std::string_view`, and can be constructed from `const std::string&`/`std::string_view` (also `std::string`).

```cpp
template <>
struct string_like<java::lang::String> : std::true_type {}
```

While I haven't found any issues specifically requesting support for non-`std` string types, I suppose #175 is a related issue. The intention is to allow custom string types to be usable with the library, however it must be opt-in (not simply inferred), to prevent things like `std::filesystem::path` from being detected as "string-like" (i.e. `std::is_convertible_v<std::filesystem::path, std::string_view>`).

**Pre-merge checklist**

    -   [x] Clang 8 or higher
    -   [x] GCC 8 or higher


include/toml++/impl/forward_declarations.hpp
```diff
@@ -379,6 +379,37 @@ TOML_NAMESPACE_START // abi namespace
...
+	/// 		 - to be usable as a retrieval target, it must be constructible from `std::string_view`
+	///
+	/// 		Types satisfying only one of the two are supported in that direction only; attempting to use one
+	/// 		in the unsupported direction is a compile error.
+	///
+	/// \note	TOML string values are always stored internally as std::string. This customization point governs
+	/// 		conversions at the API boundary only, not the underlying storage, so retrieving a value as a
+	/// 		user-defined string type makes a copy.
+	template <typename T>
+	struct string_like : std::false_type
```
marzer: Hi, thanks for the contribution. I can see the desire for this, but to be honest I'd much prefer an inferred model, as this is inline with `nlohmann::json` and other common libraries in this domain, where they detect 'string-likes'.

The basic principle would be to detect types that:
- have public `const char* data() const` or `const char* c_str() const`
- have public `size_t length() const`
- (_optionally_) not implicitly convertible to `string_view` (to avoid ambiguity).

These sorts of checks should be pretty trivial to write with some expression SFINAE (`is_detected` or similar).

mikomikotaishi: OK, I can try and look at `nlohmann::json`'s implementation and apply it here.

